PR #17, opened 2025-07-09
[BugFix] fix test cases for ci

Louis-Law:
1. Add test configuration file for test cases
2. Fixed a bug that caused DeepSeek results to fail to parse.
3. Fix circular dependency between workflow and node

### DIFF
--- a/datus/models/deepseek_model.py
+++ b/datus/models/deepseek_model.py
@@ -284,6 +284,28 @@ class DeepSeekModel(LLMBaseModel):
                     "sql_contexts": extract_sql_contexts(result),
                 }
 
+                # Create reasoning content from the full interaction list
+                reasoning_content = None
+                if hasattr(result, "to_input_list"):
+                    try:
+                        # Pass the raw list of interactions
+                        reasoning_content = result.to_input_list()
+                    except Exception as e:
+                        logger.error(f"Error getting reasoning content list: {e}")
+                        # Fallback to a simple string representation
+                        reasoning_content = str(result.to_input_list())
+
+                logger.debug(f"Reasoning content: {reasoning_content}")
+                self._save_llm_trace(
+                    prompt=prompt,
+                    response_content=result.final_output,
+                    reasoning_content=reasoning_content,
+                )
+                return {
+                    "content": result.final_output,
+                    "sql_contexts": extract_sql_contexts(result),
+                }
+            return final_result
         except Exception as e:
             logger.error(f"Error in run_agent: {str(e)}")
             reasoning_steps.append("=== Error Occurred ===")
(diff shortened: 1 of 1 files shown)


datus/agent/workflow.py
```diff
@@ -1,11 +1,12 @@
 import time
-from typing import Any, Dict, List, Optional
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 from datus.configuration.agent_config import AgentConfig
 from datus.schemas.node_models import Context, SQLContext, SqlTask
 from datus.utils.loggings import get_logger
 
-from .node import Node
+if TYPE_CHECKING:
```
Dshadowzh, line 8: Why do we need to add TYPE_CHEKCKING here? Can you remove the recursive dependency without introduce this?

datus/tools/mcp_server.py
```diff
@@ -194,20 +194,23 @@ def get_snowflake_mcp_server(cls, database_name: str, db_config: DbConfig):
                             logger.error(f"Could not find Snowflake MCP directory: {e}")
                             return None
 
-                    mcp_server_params = {
-                        "command": "uv",
-                        "args": ["--directory", directory, "run", "mcp_snowflake_server"],
-                        "env": {
+                    mcp_server_params = MCPServerStdioParams(
```
Dshadowzh, line 197: What's the difference between the MCPServerStdioParams and normal params?

Louis-Law: This is to resolve the mypy warning error.

pyproject.toml
```diff
@@ -50,7 +50,7 @@ core = [
     "aiohttp>=3.11.16",
     "xlsxwriter>=3.2.2",
     "tiktoken>=0.9.0",
-    "openai-agents==0.0.11",
+    "openai-agents>=0.1.0",
```
Dshadowzh: Have you already tested it? I am not sure the changes of openai-agent 0.1.0

requirements.txt
```diff
@@ -31,3 +31,6 @@ anthropic==0.51.0
 snowflake-sqlalchemy==1.7.3
 duckdb-engine>=0.17.0
 uv>=0.7.19
+json-repair>=0.47.6
+mysqlclient==2.2.7
```
Dshadowzh: Why do we need two mysqlclient?

datus/models/deepseek_model.py
```diff
@@ -284,26 +284,6 @@ def default(self, obj):
                     "sql_contexts": extract_sql_contexts(result),
                 }
 
-                # Create reasoning content from the full interaction list
```
Dshadowzh: Why do you delete this?

PR #25, opened 2025-07-12
[BugFix] Fixed bugs for cli

Louis-Law:
* Fixed a bug where the datus_cli could not display too many columns.

Current situation:
image

The situation after the repair:
image

### DIFF
--- a/datus/cli/repl.py
+++ b/datus/cli/repl.py
@@ -7,7 +7,7 @@ import sys
 import threading
 from enum import Enum
 from pathlib import Path
-from typing import Tuple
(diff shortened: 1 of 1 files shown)


datus/cli/repl.py
```diff
@@ -236,6 +237,66 @@ def _cmd_list_namespaces(self):
         self.console.print(table)
         return
 
+    def _smart_display_table(
```
Dshadowzh, line 240: It‘s better to add a option to compact column to fit the screen width.
if compact_mode = False, you should still display all the columns

PR #39, opened 2025-07-17
[Enhancement] Add configuration for embedding device type

Louis-Law:


pyproject.toml
```diff
@@ -41,8 +41,8 @@ dependencies = [
     "pydantic>=2.11.7",
     "Markdown==3.8",
     # "sqllineage>=0.20.0",
-    "lancedb==0.18.0",
-    "pylance==0.22.0",
+    "lancedb>=0.18.0",
```
Dshadowzh: Why do you want to upgrade the dependency?

PR #50, opened 2025-07-22
[Enhancement] Optimize  db_tools and storage configurations

Louis-Law:
1. Add method to get metadata in db_tools, including get_databases and get_schemas.
2. Optimized timing of database initialization, and exception messages
3. Fixed the bug that bootstrap-kb override mode would still report a schema mismatch.
4. Optimizing OpenAI's embedding configuration

### DIFF
--- a/datus/tools/db_tools/base.py
+++ b/datus/tools/db_tools/base.py
@@ -4,7 +4,6 @@ from typing import Any, Dict, Iterator, List, Literal, Optional, Tuple
 from pyarrow import Table as ArrowTable
 
 from datus.schemas.node_models import ExecuteSQLInput, ExecuteSQLResult
-from datus.utils.exceptions import DatusException, ErrorCode
 from datus.utils.sql_utils import metadata_identifier
 
 
@@ -116,12 +115,14 @@ class BaseSqlConnector(ABC):
         """
         raise NotImplementedError
 
-    def _reset_filter_tables(self, tables: Optional[List[str]] = None, **kwargs) -> List[str]:
+    def _reset_filter_tables(
+        self, tables: Optional[List[str]] = None, catalog_name: str = "", database_name: str = "", schema_name: str = ""
+    ) -> List[str]:
         filter_tables = []
         if tables:
-            catalog_name = kwargs.get("catalog_name", "")
-            database_name = kwargs.get("database_name", "")
-            schema_name = kwargs.get("schema_name", "")
+            catalog_name = catalog_name or self.catalog_name
+            database_name = database_name or self.database_name
+            schema_name = schema_name or self.schema_name
             for table_name in tables:
                 filter_tables.append(
                     self.full_name(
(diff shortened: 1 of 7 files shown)


conf/agent.yml.example
```diff
@@ -115,6 +115,7 @@ agent:
       model_name: text-embedding-v3
       dim_size: 1024
       batch_size: 10 # Aliyun only supports no more than 10
+      target_model: google # target model in agent.models, if not set, the agent's target configuration will be used
```
Dshadowzh: What does google means here?

datus/tools/db_tools/base.py
```diff
@@ -133,11 +140,29 @@ def get_views_with_ddl(self, **kwargs) -> List[Dict[str, str]]:
     def get_tables(self, **kwargs) -> List[str]:
         """
         Get all table names from the database.
-        Namespace parameters (such as catalog_name, database_name, schema_name)
+        Parameters contains catalog_name, database_name and schema_name
         should be passed via kwargs and handled by subclasses as needed.
         """
         raise NotImplementedError
 
+    def switch_context(self, **kwargs):
```
Dshadowzh: why do you use another do_switch_context here. The other method don't have such exception handling?

Dshadowzh: And it's better don't use **kwargs here. 
We want input of this function to be clear enough for LLM, because the more clear for parameter, the more accurate for llm tool use.

datus/tools/db_tools/db_manager.py
```diff
@@ -40,107 +42,143 @@ def gen_uri(db_config: DbConfig) -> str:
...
-    def get_conn(self, name: str, db_type: str, db_name: str = "") -> BaseSqlConnector:
-        current_name = db_config_name(name, db_type, db_name)
-        if current_name not in self._db_configs:
-            namespace_configs = []
-            for k, v in self._db_configs.items():
-                if k.startswith(f"{name}::"):
-                    namespace_configs.append(v)
-            if len(namespace_configs) == 1:
-                return self._get_conn(current_name, namespace_configs[0])
+    def __init__(self, db_configs: Dict[str, Dict[str, DbConfig]]):
```
Dshadowzh, line 45: Is this manager only for duckdb and sqlite ?  Or we can consider to put connection pool for the other db connections. 
I think duckdb & sqlite 's multi-database requirement is less important

datus/tools/db_tools/base.py
```diff
@@ -67,6 +68,12 @@ def execute_csv(self, query: str) -> ExecuteSQLResult:
     def execute_arrow_iterator(self, query: str, max_rows: int = 100) -> Iterator[ArrowTable]:
         raise NotImplementedError
 
+    def get_databases(self, catalog: str = "") -> List[str]:
```
Dshadowzh: Could you use this in cli, I think we have some remaining legacy code for sqlite in cli

datus/tools/db_tools/base.py
Louis-Law: OK

datus/tools/db_tools/base.py
Louis-Law: I'm doing it.

datus/tools/db_tools/db_manager.py
Louis-Law: It supports all databases

datus/tools/db_tools/base.py
Louis-Law: > why do you use another do_switch_context here. The other method don't have such exception handling?

Because the capabilities of this method are clear, I've done the uniform exception handling here

PR #60, opened 2025-07-25
[BugFix] Fix bug in the default configuration file

Louis-Law:


.gitignore
```diff
@@ -13,7 +13,6 @@
 /tests/llm_find_tables
 tests/data/datus*
 tests/data/metricflow_csv/*
-/conf/agent.yml
```
Dshadowzh: Why do you want to remove this?

Louis-Law: It looks like a mishap and I will revert it

PR #69, opened 2025-07-31
[Enhancement] set extra env variable for mcp-metricflow server

linghb:
set extra env variables for mcp-metricflow server so that metricflow can get them in prior and execute the cmd

### DIFF
--- a/datus/tools/mcp_server.py
+++ b/datus/tools/mcp_server.py
@@ -396,9 +396,9 @@ class MCPServer:
                         env_settings["MF_DWH_SCHEMA"] = db_config.schema
                         env_settings["MF_DWH_DIALECT"] = db_config.type
                         env_settings["MF_DWH_DB"] = str(Path(db_config.uri).expanduser())
-                    elif db_config.type == DBType.STARROCKS.value:
+                    elif db_config.type == DBType.STARROCKS:
                         env_settings["MF_DWH_SCHEMA"] = db_config.schema
-                        env_settings["MF_DWH_DIALECT"] = DBType.MYSQL.value
+                        env_settings["MF_DWH_DIALECT"] = DBType.MYSQL
                         env_settings["MF_DWH_DB"] = str(Path(db_config.uri).expanduser())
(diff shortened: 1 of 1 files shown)


datus/tools/mcp_server.py
```diff
@@ -405,6 +405,59 @@ def get_metricflow_mcp_server(cls):
                     )
         return cls._metricflow_mcp_server
 
+    @classmethod
+    def get_metricflow_with_dbinfo_mcp_server(cls, database_name: str, db_config: DbConfig):
```
stukid: why not implement this in get_metricflow_mcp_server()?

datus/tools/mcp_server.py
```diff
@@ -405,6 +405,59 @@ def get_metricflow_mcp_server(cls):
...
+                            return None
+                    logger.info(f"Using MetricFlow MCP server with directory: {directory}")
+
+                    env_settings = {
+                        "MF_MODEL_PATH": os.getenv("FILESYSTEM_MCP_DIRECTORY", "/tmp"),
+                        "MF_PATH": mf_path,
+                        "MF_PROJECT_DIR": mf_project_dir,
+                        "MF_VERBOSE": mf_verbose,
+                    }
+                    if db_config.type in (DBType.DUCKDB.value, DBType.SQLITE.value):
```
stukid: DBType is inherited from str, you can just use DBType.DUCKDB, no need to add .value.

datus/configuration/mcp_type.py
```diff
@@ -0,0 +1,4 @@
+class MCPType:
```
stukid: why MCPType here?

.gitignore
```diff
@@ -17,6 +17,7 @@ tests/data/metricflow_csv/*
 /save
 uv.lock
 /logs
+/datus/logs/*
```
stukid: Do we have logs dir under datus? If you run command under project root dir, I think no logs dir will be created under datus.

PR #72, opened 2025-08-01
[BugFix] Fix Windows compatibility issues

yww11111:
Fix fork process startup issue and wildcard recognition failure on Windows


tests/test_Windows_compatibility.py
Louis-Law, line 1: The python file should be named snake case

PR #81, opened 2025-08-05
[Feature]Mcp server

linghb:
1. desgin mcp.json and define the keys & values
2. parse mcp.json and initialize mcp_server_client based on mcp.json
3. maintian the mcp_server_client


datus/utils/json_utils.py
```diff
@@ -256,3 +257,27 @@ def load_jsonl_dict(file_path, key_field: str = "instance_id") -> Dict[str, Dict
             item = json.loads(line)
             data[item[key_field]] = item
     return data
+
+
+def resolve_json_env(value: str) -> str:
```
Louis-Law, line 262: There is already a similar implementation in agent_config.py, can it be used directly?

PR #101, opened 2025-08-13
[Enhancement] Optimizing the loading of embedded models for lazy loading

Louis-Law:
* Optimizing the loading of embedded models for lazy loading.
* Adjusting the way configuration is loaded and data is queried in integration tests.


tests/test_embedding_lazy_loading.py
Dshadowzh, line 1: It seems like not a integration tests? I think we should put these into tests/ 
Adn it's not the pytest specifcation

PR #123, opened 2025-08-23
[Feature]add date parser node

stukid:
1. add date parser node, for better relative date translation by LLM
2. split prompts template into Chinese and English versions
3. will load template from ~/.datus/template first

### DIFF
--- a/datus/models/mcp_utils.py
+++ b/datus/models/mcp_utils.py
@@ -81,4 +81,11 @@ async def multiple_mcp_servers(mcp_servers: Dict[str, Any]):
 
     finally:
         logger.debug("Cleaning up all MCP servers via AsyncExitStack")
-        await stack.aclose()
+        try:
+            await stack.aclose()
+        except RuntimeError as e:
+            if "Attempted to exit cancel scope in a different task than it was entered in" in str(e):
+                # This is a known anyio issue that can be safely ignored during cleanup
+                logger.debug("Suppressed cancel scope error during MCP server cleanup")
+            else:
+                raise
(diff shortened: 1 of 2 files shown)


datus/agent/node/date_parser_node.py
```diff
@@ -0,0 +1,192 @@
...
+                success=True, extracted_dates=extracted_dates, enriched_task=enriched_task, date_context=date_context
+            )
+
+        except Exception as e:
+            logger.error(f"Date parsing execution error: {str(e)}")
+            return DateParserResult(
+                success=False, error=str(e), extracted_dates=[], enriched_task=self.input.sql_task, date_context=""
+            )
+
+    async def _date_parsing_stream(
```
Dshadowzh: Have you tested this stream execution. I don't think we need a _date_parsing_stream in this node.

datus/agent/node/date_parser_node.py
```diff
@@ -0,0 +1,192 @@
...
+                success=False,
+                error="Date parsing model not provided",
+                extracted_dates=[],
+                enriched_task=self.input.sql_task,
+                date_context="",
+            )
+
+        try:
+            # Create date parsing tool
+            date_tool = DateParsingTool(self.model)
```
Dshadowzh: Can we remove this DateParsingTool, just implement this logic inside node?

datus/agent/node/date_parser_node.py
stukid: done

datus/agent/node/date_parser_node.py
stukid: done

PR #150, opened 2025-09-05
[Enhancement] Optimized local FunctionTool calls

Louis-Law:
* Optimized ChatNode's system prompt words
* Adjusted descriptions of some tool methods


datus/prompts/prompt_templates/chat_system_1.1.j2
Dshadowzh, line 1: Use chat_system_0.9.j2 first

PR #152, opened 2025-09-06
[Enhancement] Added @ fuzzy matches and history SQL to the CLI's chat mode

Louis-Law:


conf/.mcp.json
Dshadowzh, line 1: remove this

PR #171, opened 2025-09-16
[Feature] Web chatbot

Dshadowzh:
Start web 
` python -m datus.cli.main --web --namespace local_sqlite --port 8502 --host 0.0.0.0`

use a seperated logfile at logs/web_chatbot.xxx.log

image
image
image

### DIFF
--- a/datus/utils/loggings.py
+++ b/datus/utils/loggings.py
@@ -211,15 +211,6 @@ def setup_web_chatbot_logging(debug=False, log_dir="logs"):
     return structlog.get_logger("web_chatbot")
 
 
-def get_web_chatbot_logger() -> structlog.BoundLogger:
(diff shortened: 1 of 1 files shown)


datus/cli/main.py
```diff
@@ -92,12 +113,73 @@ def run(self):
...
-        cli.run()
+        # Check if web interface is requested
+        if args.web:
+            self._run_web_interface(args)
+        else:
+            # Initialize and run CLI
+            cli = DatusCLI(args)
+            cli.run()
+
+    def _run_web_interface(self, args):
```
Louis-Law, line 126: Would it be better to move this method into `web_chatbot.py`?

Dshadowzh: ok

PR #187, opened 2025-09-29
[BugFix]fix agentic nodes model config

stukid:
fix the bug that agentic nodes model name not works

### DIFF
--- a/datus/agent/node/agentic_node.py
+++ b/datus/agent/node/agentic_node.py
@@ -63,15 +63,7 @@ class AgenticNode(ABC):
         # Initialize the model using agent config
         if agent_config:
             model_name = self.node_config.get("model")
-            if not model_name:
-                nodes_config = agent_config.nodes if hasattr(agent_config, "nodes") else {}
-                node_name = self.get_node_name()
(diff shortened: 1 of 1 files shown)


datus/agent/node/agentic_node.py
```diff
@@ -62,17 +62,14 @@ def __init__(
...
-            node_name = self.get_node_name()
-
-            # Check for node-specific model configuration
-            if node_name in nodes_config:
-                node_config = nodes_config[node_name]
-                if hasattr(node_config, "model"):
-                    model_name = node_config.model
+            model_name = self.node_config.get("model")
+            if not model_name:
+                nodes_config = agent_config.nodes if hasattr(agent_config, "nodes") else {}
```
Dshadowzh: Using agentic_nodes.model first, I think we needn't fallback to nodes, is this for compactbility?

stukid: > model_name = self.node_config.get("model")

This is from agentic_node.model,  the rest is to keep the same as before, fall back to nodes, and then fall back to default model

PR #196, opened 2025-10-04
[BugFix]fix !gen_metrics error

stukid:
## Summary by CodeRabbit

- New Features
  - CLI now prompts for a SQL query (defaults to your last query) when generating metrics.
  - SQL identifiers containing hyphens are correctly quoted so queries parse reliably.

- Bug Fixes
  - Prevents crashes in environments with an active event loop by running async tasks in a background thread when needed.
  - More consistent error propagation and clearer logging when semantic model generation fails.

- Chores
  - Generate metrics input simplified to use sql_query; CLI no longer sends database_type or table_schemas.

### DIFF
--- a/datus/agent/node/generate_metrics_node.py
+++ b/datus/agent/node/generate_metrics_node.py
@@ -246,7 +246,6 @@ class GenerateMetricsNode(Node):
             # Check and generate semantic models if needed
             db_type = self.input.sql_task.database_type
             sql_query = quote_hyphenated_identifiers(self.input.sql_query, db_type)
-
             table_names = extract_table_names(sql_query, self.input.sql_task.database_type)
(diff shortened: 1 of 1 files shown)


datus/agent/node/generate_metrics_node.py
```diff
@@ -190,7 +244,10 @@ def _generate_metrics(self) -> GenerateMetricsResult:
 
         try:
             # Check and generate semantic models if needed
-            table_names = extract_table_names(self.input.sql_query, self.input.sql_task.database_type)
+            db_type = self.input.sql_task.database_type
+            sql_query = quote_hyphenated_identifiers(self.input.sql_query, db_type)
+
+            table_names = extract_table_names(sql_query, self.input.sql_task.database_type)
```
stukid, line 249: fixed

PR #204, opened 2025-10-07
[Refactor] Optimize ContextSearchTools and Add new CliChatAgenticNode

Dshadowzh:
* CliChatAgenticNode is inherent from GenSQLAgenticNode, and ready to substitute ChatAgenticNode
* Move some logic to AgenticNode such as mcp configuration & workspace root & plan mode
* CliChatAgenticNode & GenSQLAgenticNode have their own default tools sets

## Summary by CodeRabbit

* **New Features**
  * New CLI chat node with optional plan mode, domain/layer context, expanded chat input/result fields, and session labeling updates.
  * Semantic table search (search_table) plus domain/layer navigation tools (list_domains, list_layers_by_domain, list_items) and retrieval tools for metrics and reference SQL.
  * New metric/sql-history query APIs to browse domains, layers, and item names.

* **Improvements**
  * Streamlined prompts/output format (json sql + output) and consistent token reporting.
  * Unified CLI/chat flows, more robust streaming handling, and updated action/history displays.

### DIFF
--- a/datus/cli/agent_commands.py
+++ b/datus/cli/agent_commands.py
@@ -393,7 +393,7 @@ class AgentCommands:
             # Note: search_table is now in DBFuncTool, not ContextSearchTools
             from datus.tools.tools import db_function_tool_instance
 
-            db_tool = db_function_tool_instance(self.agent_config)
+            db_tool = db_function_tool_instance(self.cli.agent_config)
             result = db_tool.search_table(
                 query_text=input_text,
                 catalog=catalog_name,


datus/storage/metric/store.py
```diff
@@ -189,6 +197,93 @@ def get_semantic_model_size(self):
...
+
+        Returns:
+            List of unique domain names
+        """
+        self.metric_storage._ensure_table_ready()
+
+        MAX_DOMAINS = 1000
+        search_result = self.metric_storage.table.search().select(["domain"]).limit(MAX_DOMAINS).to_list()
+
+        if len(search_result) >= MAX_DOMAINS:
```
Louis-Law, line 191: Is this check redundant? The `limit` has already been called to restrict the maximum number of returns

Dshadowzh: if return MAX_DOMAINS, it print a warning log only.

PR #234, opened 2025-10-18
[BugFix]fix sqlite uri

stukid:
1. expand usr for home directory of sqlite uri
2. check uri if exists before testing connectivity
3. delete wrong config from agent.yml if test failed

## Summary by CodeRabbit

* **Bug Fixes**
  * Database connectivity failures during namespace setup now automatically remove invalid namespace entries to improve recovery.
  * Improved database connection validation: URIs are expanded before testing and SQLite checks verify the database file exists before attempting a connection.

* **Configuration**
  * Updated LLM model configuration.

### DIFF
--- a/datus/cli/interactive_init.py
+++ b/datus/cli/interactive_init.py
@@ -481,14 +481,20 @@ class InteractiveInit:
                 )
             else:
                 # For URI-based connectors (sqlite, duckdb, postgresql)
-                uri = os.path.expanduser(db_config_data.get("uri", ""))
+                uri = db_config_data.get("uri", "")
 
-                # For SQLite, verify the file exists before testing connection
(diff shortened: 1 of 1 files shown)


datus/cli/interactive_init.py
```diff
@@ -477,9 +481,18 @@ def _test_db_connectivity(self) -> tuple[bool, str]:
                 )
             else:
                 # For URI-based connectors (sqlite, duckdb, postgresql)
+                uri = os.path.expanduser(db_config_data.get("uri", ""))
```
Louis-Law: This check should be placed after `uri.replace(“sqlite:///”, “”)`.

stukid: fixed

PR #502, opened 2026-03-31
[Feature] Add AWS Bedrock support via LiteLLM (LLM + Embedding)

yhyyz:
## Summary

Add AWS Bedrock support for both LLM inference (Claude via Bedrock) and embedding models (Titan Embed, Cohere, etc.) through the LiteLLM unified interface.

### Key Changes

- **LiteLLM Embedding Provider**: New `LiteLLMEmbeddings` class with model-aware text truncation via `litellm.get_model_info()` — automatically adapts to each model's token limit instead of hardcoding
- **ClaudeModel Bedrock Adaptation**: Bedrock models bypass API key validation and native Anthropic client initialization (uses IAM role auth via boto3). Added `BEDROCK_MODEL_PREFIX` constant and `anthropic_client is None` safety guard
- **LanceDB Integration**: `_LanceLiteLLMAdapter` registered with LanceDB's embedding function system for proper serialization/deserialization
- **EmbeddingModel Factory**: Added `LITELLM` provider branch to `init_model()`, following the existing OpenAI/FastEmbed pattern

### Files Changed (8 files)

| File | Change |
|------|--------|
| `datus/utils/constants.py` | +`LITELLM` enum, +`BEDROCK_MODEL_PREFIX` constant |
| `datus/storage/embedding_litellm.py` | **NEW** — LiteLLM embedding with model-aware truncation |
| `datus/storage/embedding_models.py` | +LITELLM provider branch in `init_model()` |
| `datus/storage/vector/lance_backend.py` | +`_LanceLiteLLMAdapter` + `_wrap_embedding()` LiteLLM branch |
| `datus/models/claude_model.py` | Bedrock detection via constant + `anthropic_client` guard |
| `tests/unit_tests/storage/test_embedding_litellm.py` | **NEW** — 18 unit tests |
| `tests/unit_tests/models/test_claude_model.py` | +6 Bedrock-specific tests |
| `tests/unit_tests/storage/vector/test_lance_backend.py` | +1 LiteLLM wrapping test |

### Truncation Strategy

Replaced hardcoded `MAX_CHARS=12000` with model-aware approach:
- Uses `litellm.get_model_info(model)["max_input_tokens"]` to auto-detect each model's token limit
- Conservative char estimation: `max_input_tokens * 3` characters (handles CJK text)
- Graceful fallback to 8191 tokens (OpenAI default) if model info unavailable
- Per-text failure returns `None` instead of crashing the entire batch

### Configuration Example

```yaml
agent:
  target: bedrock_opus
  models:
    bedrock_opus:
      type: claude
      api_key: "bedrock-iam-auth"  # placeholder, IAM role used
      model: bedrock/us.anthropic.claude-opus-4-6-v1
  storage:
    database:
      registry_name: litellm
      model_name: bedrock/amazon.titan-embed-text-v2:0
      dim_size: 1024
```

### Testing

- 25 new unit tests (CI-tier, all mocked, zero external dependencies)
- 7762 existing tests pass with no regressions
- `ruff format` + `ruff check` clean

## Summary by CodeRabbit

* **New Features**
  * Added support for AWS Bedrock-targeted Claude models
  * Added LiteLLM as an embedding provider with automatic dimensionality and token limit detection

* **Improvements**
  * Enhanced error handling to prevent native Anthropic API usage with Bedrock models with clear error guidance

### DIFF
--- a/datus/storage/embedding_litellm.py
+++ b/datus/storage/embedding_litellm.py
@@ -1,0 +1,113 @@
+# Copyright 2025-present DatusAI, Inc.
+# Licensed under the Apache License, Version 2.0.
+# See http://www.apache.org/licenses/LICENSE-2.0 for details.
+
+from __future__ import annotations
+
+from functools import cached_property
+from typing import TYPE_CHECKING, List, Optional, Union
+
+import litellm
+from datus_storage_base.vector.base import EmbeddingFunction
+from pydantic import BaseModel
+
+from datus.utils.loggings import get_logger
+
+logger = get_logger(__name__)
+
+if TYPE_CHECKING:
+    import numpy as np
+
+
+class LiteLLMEmbeddings(BaseModel, EmbeddingFunction):
+    """An embedding function that uses the LiteLLM API.
+
+    Supports any embedding model that LiteLLM supports, including
+    AWS Bedrock Titan, Cohere, and other providers.
+
+    Example model names:
+        - "bedrock/amazon.titan-embed-text-v2:0"
+        - "bedrock/cohere.embed-english-v3"
+    """
+
+    name: str
+    dim: Optional[int] = None
+
+    @classmethod
+    def create(cls, **kwargs) -> "LiteLLMEmbeddings":
+        """Create a new instance with the given parameters."""
+        return cls(**kwargs)
+
+    def __setattr__(self, name: str, value: object) -> None:
+        super().__setattr__(name, value)
+        if name in {"name", "dim"}:
+            self.__dict__.pop("_ndims", None)
+            self.__dict__.pop("_max_input_tokens", None)
+
+    def ndims(self) -> int:
+        return self._ndims
+
+    @staticmethod
+    def sensitive_keys() -> list[str]:
+        return []
+
+    @cached_property
+    def _ndims(self) -> int:
+        if self.dim is not None:
+            return self.dim
+        try:
+            resp = litellm.embedding(model=self.name, input=["test"])
+            return len(resp.data[0]["embedding"])
+        except Exception as e:
+            logger.error(f"Failed to auto-detect embedding dimensions for model '{self.name}': {e}")
+            raise
+
+    @cached_property
+    def _max_input_tokens(self) -> int:
+        """Auto-detect the maximum input tokens for this embedding model."""
+        try:
+            model_info = litellm.get_model_info(self.name)
+            max_tokens = model_info.get("max_input_tokens")
+            if max_tokens and isinstance(max_tokens, int) and max_tokens > 0:
+                logger.debug(f"Model {self.name}: max_input_tokens={max_tokens}")
+                return max_tokens
+        except Exception as e:
+            logger.warning(f"Failed to get model info for '{self.name}': {e}. Using default 8191 tokens.")
+        return 8191  # OpenAI default fallback
+
+    def _truncate(self, text: str) -> str:
+        """Truncate text to fit within the model's token limit.
+
+        Uses a conservative estimate of 3 characters per token.
+        """
+        max_chars = self._max_input_tokens * 3
+        if len(text) <= max_chars:
+            return text
+        logger.warning(
+            f"Truncating text from {len(text)} to {max_chars} chars "
+            f"(~{self._max_input_tokens} tokens) for model {self.name}"
+        )
+        return text[:max_chars]
+
+    def generate_embeddings(self, texts: Union[List[str], "np.ndarray"]) -> List["np.array"]:
+        import numpy as np
+
+        valid_texts = []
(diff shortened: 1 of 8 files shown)


datus/storage/embedding_litellm.py
```diff
@@ -0,0 +1,113 @@
...
+        if not valid_texts:
+            return [None] * len(texts)
+
+        valid_embeddings: dict = {}
+        for i, (text, orig_idx) in enumerate(zip(valid_texts, valid_indices)):
+            try:
+                resp = litellm.embedding(model=self.name, input=[text])
+                valid_embeddings[orig_idx] = np.array(resp.data[0]["embedding"])
+            except Exception as e:
+                logger.error(f"LiteLLM embedding failed for text #{i} (len={len(text)}): {e}")
```
stukid, line 111: **Performance issue: one API call per text instead of batching.**

Currently each text triggers a separate `litellm.embedding()` call. For knowledge-base ingestion with hundreds or thousands of chunks, this means N sequential HTTP round-trips — very slow and wasteful.

`litellm.embedding()` natively supports batched input (`input=["text1", "text2", ...]`), so the simplest fix is:

```python
# Batch call first, fall back to per-text on failure
try:
    resp = litellm.embedding(model=self.name, input=valid_texts)
    for i, item in enumerate(resp.data):
        valid_embeddings[valid_indices[i]] = np.array(item["embedding"])
except Exception:
    # Fallback: per-text calls so one bad input doesn't kill the batch
    for i, (text, orig_idx) in enumerate(zip(valid_texts, valid_indices)):
        try:
            resp = litellm.embedding(model=self.name, input=[text])
            valid_embeddings[orig_idx] = np.array(resp.data[0]["embedding"])
        except Exception as e:
            logger.error(f"LiteLLM embedding failed for text #{i}: {e}")
```

This keeps the per-text error isolation you already have while making the happy path ~Nx faster.

datus/storage/vector/lance_backend.py
```diff
@@ -93,6 +93,28 @@ def _get_impl(self):
         return self._impl
 
 
+@register("litellm")
+class _LanceLiteLLMAdapter(TextEmbeddingFunction):
+    """LanceDB adapter for LiteLLM embedding models (Bedrock, etc.)."""
+
+    name: str = "text-embedding-ada-002"
```
stukid, line 100: **Misleading default value — `"text-embedding-ada-002"` is an OpenAI model name.**

This default was likely copied from `_LanceOpenAIAdapter`. For a LiteLLM adapter (primarily targeting Bedrock), this is confusing: if a caller forgets to pass `name`, it would silently try to call OpenAI's embedding endpoint instead of failing fast.

Suggestion:
```python
name: str = ""  # must be provided (e.g. "bedrock/amazon.titan-embed-text-v2:0")
```

An empty string will cause `litellm.embedding()` to fail immediately with a clear error, rather than routing to an unintended provider.

datus/storage/embedding_models.py
```diff
@@ -191,6 +191,18 @@ def init_model(self):
             # check if the model is initialized
             self._model.generate_embeddings(["foo"])
             logger.debug(f"Model {self.registry_name}/{self.model_name} initialized successfully")
+
+        elif self.registry_name == EmbeddingProvider.LITELLM:
+            logger.debug(f"Initializing model {self.registry_name}/{self.model_name}")
+            from datus.storage.embedding_litellm import LiteLLMEmbeddings
+
+            self._model = LiteLLMEmbeddings.create(
```
stukid, line 199: This new `LITELLM` branch only passes `name` and `dim` into `LiteLLMEmbeddings`. It drops the per-model config that `EmbeddingModel` already carries in `openai_config` (for example `api_key` / `base_url` via `target_model`). As written, Bedrock via ambient IAM may work, but other LiteLLM-backed providers still depend on environment credentials instead of the configured target model, so the integration is incomplete.

datus/storage/embedding_litellm.py
```diff
@@ -0,0 +1,113 @@
...
+                valid_texts.append(self._truncate(text))
+                valid_indices.append(idx)
+
+        if not valid_texts:
+            return [None] * len(texts)
+
+        valid_embeddings: dict = {}
+        for i, (text, orig_idx) in enumerate(zip(valid_texts, valid_indices)):
+            try:
+                resp = litellm.embedding(model=self.name, input=[text])
```
stukid, line 108: `dim` is used as the declared embedding size (`ndims()` returns it directly), but `generate_embeddings()` does not pass `dimensions=self.dim` to LiteLLM here. That can leave Datus/LanceDB believing the vectors are one size while the provider actually returns another. If configurable dimensions are part of the contract for this provider, the request path needs to enforce it, or the implementation should stop treating `dim` as authoritative.

PR #685, opened 2026-04-28
[BugFix] Add json_repair fallback for malformed LLM tool call arguments

FrankSen:
## Why

Some LLM models (e.g. GLM) produce invalid JSON in tool call arguments, such as unquoted string values like `{"keywords": PERCENTILE function, "platform": "starrocks"}`. This causes tool calls to fail with a hard JSON parse error, breaking the agent workflow.

## Solution

Add a `json_repair` fallback in both `trans_to_function_tool` (base.py) and `SubAgentTaskTool._invoke` (sub_agent_task_tool.py). When `json.loads` fails, attempt repair via `json_repair.loads` before returning a hard error. A warning is logged when repair succeeds.

## Test Cases

- `test_malformed_json_repaired_by_json_repair` — verifies GLM-style malformed JSON is repaired and tool call succeeds
- `test_malformed_json_repair_logs_warning` — verifies warning is logged on repair
- `test_invalid_json_returns_error` — updated to use truly unrepairable input
- `TestTaskToolJsonRepair::test_malformed_json_repaired` — verifies repair in SubAgentTaskTool
- `TestTaskToolJsonRepair::test_invalid_json_still_returns_error` — verifies unrepairable JSON still errors

## Summary by CodeRabbit

* **Bug Fixes**
  * Centralized argument parsing with improved JSON handling: auto-repairs malformed JSON when possible (logs a warning), preserves prior "Invalid JSON" behavior, adds explicit "missing required fields" and truncation hints, and halts execution early on parse/validation failures. The task tool now enforces required arguments.

* **Tests**
  * Expanded unit tests covering repairable vs unrecoverable JSON, repair warnings, missing-required-fields, truncated inputs, and task-invocation outcomes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Datus-ai/Datus-agent/pull/685?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/datus/tools/func_tool/base.py
+++ b/datus/tools/func_tool/base.py
@@ -7,6 +7,7 @@ import inspect
 import json
 from typing import Any, Callable, Dict, List, Optional
 
+import json_repair
 from agents import FunctionTool, function_tool
 from pydantic import BaseModel, Field
 
@@ -127,18 +128,29 @@ def trans_to_function_tool(bound_method: Callable, *, strict_mode: bool = True) 
                 else:
                     args_dict = {}
             except (TypeError, json.JSONDecodeError) as e:
-                args_len = len(args_str) if isinstance(args_str, str) else 0
-                truncated_hint = ""
-                if isinstance(args_str, str) and args_len > 0:
-                    # Check if it looks like truncated output (no closing brace)
-                    stripped = args_str.rstrip()
-                    if not stripped.endswith("}") and not stripped.endswith("]"):
-                        truncated_hint = " Output appears truncated — likely hit model max_output_tokens limit."
-                return {
-                    "success": 0,
-                    "error": f"Invalid JSON arguments ({e}). Args length: {args_len} chars.{truncated_hint}",
-                    "result": None,
-                }
+                if isinstance(args_str, str) and args_str.strip():
+                    try:
+                        args_dict = json_repair.loads(args_str)
+                        if not isinstance(args_dict, dict):
+                            raise ValueError("Repaired result is not a dict")
+                        logger.warning(f"Repaired malformed JSON arguments for tool '{method_to_call.__name__}': {e}")
+                    except Exception:
+                        args_len = len(args_str)
+                        truncated_hint = ""
+                        stripped = args_str.rstrip()
+                        if not stripped.endswith("}") and not stripped.endswith("]"):
+                            truncated_hint = " Output appears truncated — likely hit model max_output_tokens limit."
+                        return {
+                            "success": 0,
+                            "error": f"Invalid JSON arguments ({e}). Args length: {args_len} chars.{truncated_hint}",
+                            "result": None,
+                        }
+                else:
+                    return {
+                        "success": 0,
+                        "error": f"Invalid JSON arguments ({e})",
+                        "result": None,
+                    }
 
             # Call sync or async bound methods transparently
             if inspect.ismethod(method_to_call):
--- a/datus/tools/func_tool/base.py
+++ b/datus/tools/func_tool/base.py
@@ -92,6 +92,75 @@ class FuncToolListResult(BaseModel):
     )
 
 
+def parse_tool_args(
+    args_str: Any,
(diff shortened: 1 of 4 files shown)


datus/tools/func_tool/base.py
```diff
@@ -119,18 +120,29 @@ async def final_invoker(tool_ctx, args_str) -> dict:
...
-                    if not stripped.endswith("}") and not stripped.endswith("]"):
-                        truncated_hint = " Output appears truncated — likely hit model max_output_tokens limit."
-                return {
-                    "success": 0,
-                    "error": f"Invalid JSON arguments ({e}). Args length: {args_len} chars.{truncated_hint}",
-                    "result": None,
-                }
+                if isinstance(args_str, str) and args_str.strip():
+                    try:
+                        args_dict = json_repair.loads(args_str)
```
stukid: This fallback is too permissive. `json_repair.loads()` can turn malformed or truncated inputs such as `not-valid-json{` into `{}` or `{"x":` into `{"x": ""}`. Because the wrapper then proceeds to call the real tool, a parse failure can become either an unhandled `TypeError` from missing required parameters or a false successful invocation with empty values.

Suggested fix: keep the original structured error path for clearly truncated/non-object inputs, and make repair a conservative step:
1. try `json.loads()` first;
2. only attempt repair for strings that look like a complete JSON object;
3. require the repaired value to be a dict;
4. validate required parameters against the function signature/schema before invoking the tool;
5. preserve the existing args length/truncation diagnostics when validation fails.

Please also add regression tests for `not-valid-json{`, `{"x":`, `{}`, `[]`, and a missing required parameter so these cases never reach the underlying tool.

datus/tools/func_tool/sub_agent_task_tool.py
```diff
@@ -245,8 +246,19 @@ def available_tools(self) -> List[Tool]:
         async def _invoke(_tool_ctx, args_str) -> dict:
             try:
                 args = json.loads(args_str) if isinstance(args_str, str) else dict(args_str or {})
-            except (TypeError, json.JSONDecodeError):
-                return FuncToolResult(success=0, error="Invalid JSON arguments for task tool").model_dump()
+            except (TypeError, json.JSONDecodeError) as e:
+                if isinstance(args_str, str) and args_str.strip():
+                    try:
+                        args = json_repair.loads(args_str)
```
stukid: The same repair path is needed here, but it should not directly trust `json_repair.loads()` output. For the task tool, repaired args still need to satisfy the task schema (`type`, `prompt`, and `description`) before `self.task()` is invoked; otherwise malformed input can silently become missing/empty task arguments and produce a misleading downstream error.

Suggested fix: extract a shared helper from the generic tool wrapper, for example `parse_tool_args(args_str, required_fields=..., tool_name=...)`, and use it in both `trans_to_function_tool()` and this task tool. The helper should only accept repaired dicts that pass required-field validation and should return the same structured invalid-JSON error shape for truncated or unrecoverable input.

FrankSen:
@stukid Thanks for the review! I've addressed your feedback in the latest commits:

1. **Extracted shared `parse_tool_args()` helper** (`base.py`) — centralizes JSON parsing, `json_repair` fallback, and required-field validation in one place.
2. **Required-field validation after repair** — `json_repair` can silently produce empty dicts (e.g. `"not-valid-json{"` → `{}`) or incomplete dicts (e.g. `'{"x":'` → `{"x": ""}`). Now these are caught and return a clear error with the missing field names.
3. **Deduplicated `sub_agent_task_tool.py`** — replaced inline JSON repair logic with a single `parse_tool_args()` call; removed direct `json_repair` import.
4. **Added regression tests** — covers edge cases: empty dict, array input, truncated JSON, missing required fields, and direct `parse_tool_args` unit tests.

Ready for another look when you have time!

FrankSen:
@stukid PR has been rebased onto latest main. All CI checks passing. Ready for re-review and merge.

PR #704, opened 2026-05-06
[BugFix] Resolve agent_config.home through path_manager so tilde expands

arnozhang:
## Why

`POST /api/v1/agent/edit` (and `create`) crashes with `FileNotFoundError`
when `agent_config.home` is the production default `'~/.datus'`:

```
File "/.../datus/api/services/agent_service.py", line 197, in _save_agentic_nodes
    with open(config_path) as f:
FileNotFoundError: [Errno 2] No such file or directory: '~/.datus/agent.yml'
```

`agent_config.home` is stored as a **raw string** (default `"~/.datus"`).
The four sites in `agent_service.py` that build filesystem paths via
`Path(agent_config.home)` never expand the tilde — `pathlib.Path()`
does not call `expanduser()` — so `open()` sees the literal `~` and
aborts.

The rest of the codebase already routes through
`agent_config.path_manager.datus_home`, which is the already-expanded
resolved `Path` (via `PathManager.resolve_home → Path(home).expanduser().resolve()`).
This PR makes the agent CRUD paths agree with that convention; no new
helper introduced.

## Solution

All four `Path(agent_config.home)` usages in `datus/api/services/agent_service.py`
swapped to `agent_config.path_manager.datus_home`:

- `_save_agentic_nodes` → `agent.yml` (the actual crash site)
- `get_agent`'s `created_at` file-mtime fallback
- `_copy_prompt_template` → `{home}/template`
- `_save_prompt_template` → `{home}/template`

The fix is a one-line swap per call site, no behavior change for
configs whose `home` was already absolute (the resolve step is a no-op
for absolute paths).

## Test Cases

Two regression tests added to `tests/unit_tests/api/services/test_agent_service.py`:

- **`test_edit_existing_agent_when_home_has_tilde`** — exercises
  `create_agent` + `edit_agent` end-to-end with `agent_config.home`
  forced to a literal `'~/...'` string that does not exist on disk.
  Pre-fix this raised `FileNotFoundError`; post-fix the operation
  succeeds and the change persists under `path_manager.datus_home`.
- **`test_get_agent_uses_path_manager_when_home_has_tilde`** — covers
  the `created_at` file-mtime fallback in `get_agent` for the same
  scenario.

Gates: ruff format/check clean · 62/62 unit tests pass · audit P0=0/P1=0
· diff coverage 100%.

## Summary by CodeRabbit

* **Refactor**
  * Centralized agent configuration and template storage in the datus home directory for improved organization.
  * Enhanced timestamp handling for agent creation dates.

* **Tests**
  * Added test coverage for path handling with tilde expansion in agent home directories.

### DIFF
--- a/datus/api/services/agent_service.py
+++ b/datus/api/services/agent_service.py
@@ -193,7 +193,7 @@ def _save_agentic_nodes(agent_config: AgentConfig, nodes: dict) -> None:
     """Save agentic_nodes back to agent.yml."""
     import yaml
 
-    config_path = Path(agent_config.home) / "agent.yml"
+    config_path = agent_config.path_manager.datus_home / "agent.yml"
     with open(config_path) as f:
         raw = yaml.safe_load(f)
     raw["agentic_nodes"] = nodes
@@ -261,7 +261,7 @@ class AgentService:
         agent_type = agent.get("type", "gen_sql")
         created_at = _normalize_created_at(agent.get("created_at"))
         if not created_at:
-            created_at = _file_mtime_iso(Path(agent_config.home) / "agent.yml")
+            created_at = _file_mtime_iso(agent_config.path_manager.datus_home / "agent.yml")
 
         return Result(
             success=True,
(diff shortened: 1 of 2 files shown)


arnozhang:
Folded into #705. The ConfigurationManager refactor there supersedes the `_save_agentic_nodes` and `get_agent`-mtime fixes from this PR (the save path no longer touches `agent_config.home` at all). The remaining two sites — `_copy_prompt_template` / `_save_prompt_template` — plus the tilde regression tests are now in #705 (commit e3460faf). Closing in favor of that single PR.

PR #733, opened 2026-05-11
[BugFix] Parameterize nightly compose ports

stukid:
## Why

Nightly compose suites were binding shared self-hosted runner ports such as 8088, 8030, and 15432 directly. Those ports can already be owned by long-running runner services, causing nightly to fail before the relevant tests start.

## Solution

- Move nightly compose suites to dedicated high host ports via explicit workflow env.
- Add per-suite Docker Compose project names and cleanup so nightly only stops its own services.
- Add host-port preflight diagnostics that report the owning listener before compose startup.
- Log the effective compose port configuration for debugging.
- Serialize nightly workflow runs with a workflow-level concurrency group.

## Test Cases

- `bash -n ci/run-nightly-tests.sh`
- `git diff --check`
- `NIGHTLY_GROUP_FILTER=NoSuchNightlyGroup NIGHTLY_LOG_FILE=/tmp/datus-nightly-portparam-smoke2.log ci/run-nightly-tests.sh`
- Verified the selected self-hosted runner has the new high ports free: `18088 15433 18080 25432 23306 28123 29000 29030 28030 28080 15434 29083 21000 21002 31000 24040`
- Verified external adapter compose config resolves Superset, StarRocks, Greenplum, and Airflow to the new parameterized ports.

## Summary by CodeRabbit

* **Chores**
  * Added serialized nightly runs with non-canceling concurrency to avoid interrupting in-progress suites.
  * Per-run compose project naming to scope services and prevent cross-run collisions.
  * Exported per-service host/port environment variables so nightly suites use non-colliding ports (Superset, Airflow, Postgres, MySQL, ClickHouse, StarRocks, Trino, Greenplum, Hive, Spark).
  * Verifies host port bindability before startup and collects scoped diagnostics on failures for reliable teardown.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Datus-ai/Datus-agent/pull/733)

### DIFF
--- a/.github/workflows/run-nightly.yml
+++ b/.github/workflows/run-nightly.yml
@@ -6,6 +6,9 @@ on:
   workflow_dispatch: # Allow manual trigger
 
 concurrency:
+  # Keep nightly runs serialized because the workflow intentionally uses the
+  # stable datus-nightly-* compose project names below. Per-suite startup runs
+  # compose down first to clean leftovers from force-cancelled runs.
   group: datus-agent-nightly
   cancel-in-progress: false
 
--- a/.github/workflows/run-nightly.yml
+++ b/.github/workflows/run-nightly.yml
@@ -5,6 +5,10 @@ on:
     - cron: '0 20 * * *' # 20:00 UTC daily (offset from integration at 18:00 UTC)
   workflow_dispatch: # Allow manual trigger
 
+concurrency:
+  group: datus-agent-nightly
+  cancel-in-progress: false
+
 jobs:
   run-nightly:
     runs-on: self-hosted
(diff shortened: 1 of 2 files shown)


stukid:
Addressed the 3 CodeRabbit nitpicks:

- Split the `compose_project_slug` command substitution out of the `local` assignment in `compose_project_name`, so the command status is not hidden by the assignment.
- Updated `compose_host_port_specs` fallback values to use the new high ports, so diagnostics no longer fall back to the old conflict-prone ports.
- Added a workflow concurrency comment documenting why the stable `datus-nightly-*` compose project prefix depends on serialized nightly runs and per-suite `compose_down` cleanup.

Validation run:
- `bash -n ci/run-nightly-tests.sh`
- `git diff --check`
- `NIGHTLY_GROUP_FILTER=NoSuchNightlyGroup NIGHTLY_LOG_FILE=/tmp/datus-nightly-portparam-smoke3.log ci/run-nightly-tests.sh`

PR #843, opened 2026-05-20
[Enhancement] move Skeleton/ErrorBlock to @datus/web-artifact + use refetch

arnozhang:
## Why

The `Skeleton` / `ErrorBlock` placeholders the `gen_visual_report` and `gen_visual_dashboard` subagents emit have two long-standing problems:

1. **Cheap default look.** Generated per artifact, they end up as a single pulsing dot and a bare red text line — the first thing a reader sees while a query resolves or fails.
2. **No retry affordance.** ErrorBlock dead-ends the reader; the only "fix" was reloading the entire iframe.

Worse, both are LLM-emitted (~70 lines of JSX in every generated `render/shared/card.jsx`), so quality varies across runs and fixing the look means regenerating every existing artifact.

## Solution

This PR lands alongside [Datus-saas#390](https://github.com/Datus-ai/Datus-saas/pull/390), which moves `Skeleton` / `ErrorBlock` into `@datus/web-artifact` as first-class runtime primitives (mirroring `AnimateOnVisible` / `ChartEntryMore`) and adds `refetch` to the `useQuerySql` return value.

Prompt changes:

- **Allowed-imports table** lists `Skeleton` / `ErrorBlock` alongside the other `@datus/web-artifact` exports.
- **Canonical component shape** imports the placeholders from the runtime; `./shared/card` is left holding only `Card`.
- **New "Loading & error placeholders" section** drops the reference jsx (now ~70 lines lighter) and keeps just the prop table + wiring rules.
- **`useQuerySql` Returns doc** adds `refetch: () => void` and tells the subagent to pass it as `onRetry={refetch}` — the placeholder's `onRetry` falls back to `window.location.reload()` so the artifact still has a working recovery path if a component omits it.
- **Canonical examples and the Data-layer intro** destructure `refetch` from the hook and wire `<ErrorBlock onRetry={refetch} />` in early returns.

Net effect on the rendered prompt: **-2.6 KB / -70 lines**, plus a consistent placeholder look across every artifact (existing and future) and a per-card Retry button that doesn't reload the iframe.

### Ordering

This PR is safe to merge before #390 ships:

- If the prompt lands first, generated code imports `Skeleton` / `ErrorBlock` from `@datus/web-artifact`. The validator's allow-list already permits that bare specifier, so static validation passes; runtime resolution fails until #390 ships, which surfaces as a per-card error — not a hard crash. Practically you'd want #390 to land first.
- If #390 lands first, nothing changes for already-generated artifacts; new generations pick up the new pattern only after this PR lands.

## Test Cases

No new tests — pure prompt change. Verification:

- `jinja2` rendering of `_visual_artifact_common.j2` succeeds; the new section appears in the output and the old reference jsx (`PLACEHOLDER_KEYFRAMES`, `datus-shimmer` keyframe definitions) is gone.
- `pytest tests/unit_tests/agent/node/test_gen_visual_{dashboard,report}_agentic_node.py` — 29/29 pass (these exercise prompt loading).
- End-to-end behavior (the Retry button actually re-running the query) is covered by nightly visual-artifact regression once #390 ships.

## Summary by CodeRabbit

* **New Features**
  * Chart action menu now requires query results (data) for certain actions; those actions are disabled when data isn't provided.
  * Error states include standardized retry wiring (uses provided retry handler; falls back to page reload).

* **Documentation**
  * Updated guidance for loading/error placeholders and mandated direct imports for them.
  * Loading skeleton animation moved to card boundaries for consistent UX.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Datus-ai/Datus-agent/pull/843?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/datus/prompts/prompt_templates/_visual_artifact_common.j2
+++ b/datus/prompts/prompt_templates/_visual_artifact_common.j2
@@ -216,11 +216,15 @@ Treat each output as a polished deliverable — a small **branded analytics prod
 `Skeleton` and `ErrorBlock` are runtime-provided primitives — **import them directly from `@datus/web-artifact`**. Do not re-implement them in `render/shared/card.jsx`; every artifact shares one canonical look and gets free upgrades when the runtime ships new versions, exactly like `AnimateOnVisible` and `ChartEntryMore`.
 
 ```jsx
-import { Skeleton, ErrorBlock } from '@datus/web-artifact';
+import { useDatusArtifact, Skeleton, ErrorBlock } from '@datus/web-artifact';
 
-const { isLoading, errorMessage, data, refetch } = useQuerySql('queries/foo', params);
-if (isLoading) return <Skeleton minHeight={300} />;
-if (errorMessage) return <ErrorBlock msg={errorMessage} onRetry={refetch} />;
+function RevenueCard({ params }) {
+  const { useQuerySql } = useDatusArtifact();
+  const { isLoading, errorMessage, data, refetch } = useQuerySql('queries/foo', params);
+  if (isLoading) return <Skeleton minHeight={300} />;
+  if (errorMessage) return <ErrorBlock msg={errorMessage} onRetry={refetch} />;
+  // ... chart JSX
+}
 ```
 
 Props:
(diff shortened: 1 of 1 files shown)


arnozhang:
Updated to use the new `refetch` action on `useQuerySql` instead of the original `retryKey` state hack. The runtime change lives in Datus-saas#390 (`feat(web-artifact): expose refetch from useQuerySql`).

Ordering note: this PR is safe to merge before #390 ships. If the prompt lands first, generated components destructure `refetch` from a hook that doesn't return it yet — `refetch` is `undefined`, `ErrorBlock`'s `onRetry={undefined}` falls back to `window.location.reload()`, and Retry still works (just at iframe granularity). Once #390 is in, the same generated code automatically upgrades to per-card refetch with no further prompt change.

PR #865, opened 2026-05-22
[BugFix] Write repaired tool call arguments back to session history

### LINKED ISSUE #862: [BugFix] Repaired tool call arguments not written back to session history, causing repeated API failures
## Problem

When an LLM (e.g. GLM) returns malformed JSON in tool call arguments, the current `json_repair` fallback (introduced in PR #685) successfully parses the arguments and allows the tool to execute. However, **the repaired JSON is never written back to `raw_item.arguments`** — the original malformed string remains in the session history.

On subsequent turns, the OpenAI Agents SDK replays the full conversation history (including the malformed `arguments` string) back to the LLM API. The API then rejects the request with:

```
litellm.BadRequestError: OpenAIException - unexpected character: line 1 column 148 (char 147)
```

This causes **all subsequent requests in the same session to fail permanently**, since the corrupted history entry is never corrected.

## Root Cause

Two independent code paths handle tool call arguments:

1. **Tool execution path** (`datus/tools/func_tool/base.py:parse_tool_args`): detects malformed JSON → repairs via `json_repair` → returns valid dict → tool executes successfully
2. **Session storage path** (`datus/models/openai_compatible.py:~1091`): reads `raw_item.arguments` directly → stores in session → replayed verbatim on next API call

The repair in path (1) never propagates back to `raw_item.arguments`, so path (2) stores and replays the original malformed string.

## Reproduction

1. Configure a model that occasionally produces malformed JSON tool calls (e.g. GLM series)
2. Trigger a tool call (e.g. `ask_user`) where the model outputs invalid JSON like `{"keywords": PERCENTILE function, "platform": "starrocks"}`
3. Observe: first tool call succeeds (json_repair works), but all subsequent messages in the same session fail with `BadRequestError: unexpected character`

## Proposed Fix

In `datus/models/openai_compatible.py` around line 1091, after extracting `arguments` from `raw_item`, validate and repair before storing:

```python
arguments = getattr(raw_item, "arguments", "{}")
try:
    json.loads(arguments)
except (json.JSONDecodeError, TypeError):
    try:
        repaired = json_repair.loads(arguments)
        repaired_str = json.dumps(repaired, ensure_ascii=False)
        logger.warning(
            f"Repaired malformed tool call arguments for '{tool_name}' in session history: "
            f"original={arguments[:200]}"
        )
        if hasattr(raw_item, "arguments"):
            raw_item.arguments = repaired_str
        arguments = repaired_str
    except Exception:
        logger.error(f"Failed to repair tool call arguments for '{tool_name}': {arguments[:200]}")
```

This ensures the session always stores valid JSON, preventing cascading failures on subsequent turns.

## Related

- PR #685: Added `json_repair` fallback for tool execution (partial fix — tool runs but session is still corrupted)
- Commit `5a5861f0` (#533): Fixed tool result serialization in session history (similar class of bug)

FrankSen:
## Why

When an LLM (e.g. GLM) returns malformed JSON in tool call arguments, the `json_repair` fallback (PR #685) successfully parses them for tool execution, but **the original malformed string stays in `raw_item.arguments`**. The Agents SDK later serializes this via `to_input_item()` → `model_dump()` into session history. On subsequent turns, the corrupted history is replayed to the API, causing permanent `BadRequestError` for the entire session.

Closes #862

## Solution

In the streaming layer (`openai_compatible.py` and `codex_model.py`), reuse the existing `json.loads(arguments)` try/except block. When parsing fails, repair via `json_repair` and replace `raw_item` with a corrected Pydantic model copy (`model_copy(update={"arguments": repaired_str})`). This ensures `to_input_item()` serializes valid JSON into session storage.

- **Zero overhead for valid JSON** — repair only triggers in the existing `except` branch
- **Covers both model backends** — `OpenAICompatibleModel` (Pydantic raw_item) and `CodexModel` (dict or Pydantic raw_item)

## Test Cases

- `tests/unit_tests/models/test_openai_compatible_json_repair.py` — 6 new tests:
  - `test_valid_json_passes_through_unchanged`
  - `test_malformed_json_is_repaired` — GLM-style `PERCENTILE function` unquoted value
  - `test_model_copy_preserves_other_fields` — name/call_id unchanged after repair
  - `test_unrepairable_json_left_unchanged`
  - `test_empty_arguments_not_repaired`
  - `test_glm_style_unquoted_values_repaired` — Chinese unquoted values + array

## Summary by CodeRabbit

* **Bug Fixes**
  * Better handling of malformed JSON in streamed tool-call arguments: attempts automatic repair, updates stored/displayed arguments when repaired, and falls back to original on failure.

* **New Features**
  * Added a JSON repair utility that normalizes and validates tool-call arguments and reports when repair occurred; streaming paths now apply it.

* **Tests**
  * Added extensive unit and integration tests for valid, malformed, edge-case, and model-vs-dict repair scenarios.

* **Chores**
  * Removed a redundant local import.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3

### DIFF
--- a/datus/models/codex_model.py
+++ b/datus/models/codex_model.py
@@ -12,6 +12,7 @@ import uuid
 from contextlib import nullcontext
 from typing import Any, AsyncGenerator, Dict, List, Optional, Union
 
+import json_repair
 from agents import Agent, ModelSettings, Runner, SQLiteSession, Tool
 from agents.exceptions import MaxTurnsExceeded
 from agents.mcp import MCPServerStdio
@@ -649,7 +650,33 @@ class CodexModel(LLMBaseModel):
                                     arguments = getattr(raw_item, "arguments", "{}")
                                 if not call_id:
                                     call_id = f"tool_{uuid.uuid4().hex[:8]}"
-                                args_str = str(arguments)[:80]
+
+                                # Validate and repair malformed JSON arguments
+                                try:
+                                    json.loads(arguments)
+                                    args_str = str(arguments)[:80]
+                                except (json.JSONDecodeError, TypeError, ValueError):
+                                    try:
+                                        repaired = json_repair.loads(arguments)
+                                        repaired_str = json.dumps(repaired, ensure_ascii=False)
+                                        logger.warning(
+                                            f"Repaired malformed tool call arguments for '{tool_name}' "
+                                            f"in session history: {arguments[:200]}"
+                                        )
+                                        if isinstance(raw_item, dict):
+                                            raw_item["arguments"] = repaired_str
+                                        else:
+                                            event.item.raw_item = raw_item.model_copy(
+                                                update={"arguments": repaired_str}
+                                            )
+                                            raw_item = event.item.raw_item
+                                        arguments = repaired_str
+                                        args_str = repaired_str[:80]
+                                    except Exception:
+                                        logger.error(
+                                            f"Cannot repair tool call arguments for '{tool_name}': {arguments[:200]}"
+                                        )
+                                        args_str = str(arguments)[:80]
 
                                 temp_tool_calls[call_id] = {
                                     "tool_name": tool_name,
--- a/datus/models/codex_model.py
+++ b/datus/models/codex_model.py
@@ -12,7 +12,6 @@ import uuid
 from contextlib import nullcontext
 from typing import Any, AsyncGenerator, Dict, List, Optional, Union
 
-import json_repair
 from agents import Agent, ModelSettings, Runner, SQLiteSession, Tool
 from agents.exceptions import MaxTurnsExceeded
 from agents.mcp import MCPServerStdio
@@ -28,6 +27,7 @@ from datus.models.mcp_utils import multiple_mcp_servers
 from datus.observability.manager import get_observability_manager
 from datus.schemas.action_history import ActionHistory, ActionHistoryManager, ActionRole, ActionStatus
 from datus.utils.exceptions import DatusException, ErrorCode
+from datus.utils.json_utils import repair_tool_call_arguments
 from datus.utils.loggings import get_logger
 from datus.utils.trace_context import build_agents_run_config_kwargs, build_trace_span_attributes
 
@@ -349,7 +349,6 @@ class CodexModel(LLMBaseModel):
         Returns:
             Parsed JSON response as a dictionary
         """
-        import json
 
         self._refresh_client_token()
         input_data = self._convert_prompt_to_input(prompt)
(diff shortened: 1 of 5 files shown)


FrankSen:
Rebased onto latest `main` and addressed review feedback:

- Fixed P1 weak assertions (replaced `assert isinstance(was_repaired, bool)` and `assert "question" in parsed` with concrete value checks)
- Added edge-case tests: dict-based raw_item repair path (CodexModel), truncated JSON, trailing commas, single quotes
- Diff is now minimal (4 files, ~160 lines) — previous coverage failure was caused by stale base

@stukid Ready for re-review when you have a moment.

stukid:
## Review: the fix is correct — main gap is a regression test that pins the serialization contract

I traced the full write-back → serialization → session-storage chain and verified it end to end, so I'm confident this genuinely fixes #862 (not a cosmetic-only change):

1. `agents._run_impl.stream_step_items_to_queue` emits `RunItemStreamEvent(item=item, name="tool_called")` using the **same** `ToolCallItem` instance from `new_step_items` — i.e. the `event.item` this PR mutates is the same object that ends up in `new_items`/session items (tool-call items aren't filtered/copied).
2. `ToolCallItem.to_input_item()` reads `self.raw_item` at call time: a `dict` is returned directly; a `BaseModel` is dumped via `model_dump(exclude_unset=True)`.
3. The PR's write-back targets exactly that `raw_item` (dict in-place, or `model_copy(update={...})` reassigned to `event.item.raw_item`).
4. Verified empirically that `model_copy(update={"arguments": ...})` keeps `arguments` in `model_fields_set`, so `model_dump(exclude_unset=True)` includes the repaired value (it isn't excluded).

So the repaired JSON really does reach session history → the cascading `BadRequestError` on replay is fixed. 👍

The `repair_tool_call_arguments` helper is also clean: valid JSON passes through with zero overhead, malformed input is repaired with `json_repair`, re-dumped (`ensure_ascii=False`) and **re-validated** with `json.loads` before being returned. Covering both backends and distinguishing dict vs. Pydantic `raw_item` in CodexModel is the right call.

### What's actually missing: a regression test on the contract, not the mutation

The 6 new tests assert that `event.item.raw_item` was mutated, but none call `event.item.to_input_item()` / exercise session serialization. That leaves the load-bearing link (write-back → `to_input_item()` → stored JSON) unguarded: it's correct today, but a future SDK bump (e.g. a change to `to_input_item` or `exclude_unset` semantics) or a refactor could silently regress it with all tests still green. Please add at least one test that, after the stream, asserts `event.item.to_input_item()["arguments"]` is valid JSON (for both the dict and Pydantic paths). That pins the behavior #862 actually depends on.

### Minor / robustness (non-blocking, edge hardening)

1. **`openai_compatible.py` has no dict branch** for `raw_item` (CodexModel does). It doesn't crash today only because `getattr(raw_item, "arguments", "{}")` yields `"{}"` for a dict → repair is skipped. If a dict `raw_item` with real args ever reaches this path, its arguments are neither displayed nor repaired. Worth unifying with CodexModel or a comment asserting `raw_item` is always Pydantic here.
2. **`json_repair.loads` can return a non-object** (list/scalar) for pathological input. The helper validates "is parseable JSON" but not "is a JSON object"; tool-call arguments should be an object. Consider an `isinstance(repaired, dict)` guard (else treat as unrepairable).
3. **Empty arguments aren't normalized**: the helper returns `("{}", False)` for empty input, but the caller only uses the value when `was_repaired` is `True`, so the `"{}"` is dead — empty `""` is left as-is. Minor; either normalize empty args or return `("", False)`.
4. Worth a one-line note that repair may change argument *semantics* (e.g. unquoted `PERCENTILE function` → string `"PERCENTILE function"`). That's an acceptable trade-off (valid + session survives ≫ session-fatal), just good to make explicit in the log/comment.

### Merge requirements

- **Resolve all open review comments** (mine + CodeRabbit) before merge.
- **Rebase onto the latest `main`** before merge so this lands on current code.

Overall: approve in principle — correct fix for a real session-fatal bug. The one thing I'd want in before merge is the `to_input_item()`-level regression test; the rest is edge hardening. Happy to push the test + guards if useful.

PR #870, opened 2026-05-25
[Enhancement] Degrade gracefully when embeddings are unavailable

### LINKED ISSUE #868: [Enhancement] Gracefully degrade when embedding model is unavailable
## Background

During the investigation of #834, the user environment could not connect to Hugging Face, and the REPL log stopped at:

```text
Starting new HTTPS connection (1): huggingface.co:443
```

The default local embedding setup uses fastembed / `all-MiniLM-L6-v2`. When the model is not already cached locally, the implementation attempts to download `qdrant/all-MiniLM-L6-v2-onnx` from Hugging Face. In offline, intranet, or Hugging-Face-blocked environments, this can make REPL startup or the first interaction appear to hang without a clear user-facing error.

## Problem

This is not an LLM provider configuration issue, and REPL mode and Web mode do not use different vector-model mechanisms. More precisely:

- REPL mode preloads autocomplete/RAG context such as `@Table`, `@Metrics`, and `@Sql`, so it triggers vector store / embedding model initialization earlier.
- Web mode does not necessarily preload those context/RAG capabilities on startup, so it may appear to work at first. However, when Web chat actually uses `context_search_tools`, KB, or @-reference related capabilities, it still goes through the same embedding path.
- When the embedding model is not cached and Hugging Face is unreachable, the current product experience lacks clear error reporting and graceful feature-level degradation.

## Expected Improvements

1. REPL startup should not block basic chat/database capabilities if autocomplete or RAG preloading fails.
2. If `ContextSearchTools` or storage initialization fails, only embedding-dependent capabilities such as context search, KB search, and @ references should be disabled. DB tools should continue to work.
3. Basic database questions, such as “how many tables are in the database?”, should still be answerable through DB tools.
4. CLI and Web should show a clear message, including:
   - embedding model cache missing
   - failed to download from Hugging Face
   - context search disabled
   - suggested remediation: configure a proxy/mirror, pre-cache the model, or use an OpenAI-compatible embedding provider
5. Logs should include the actual cache path, such as `HF_HOME`, `FASTEMBED_CACHE_PATH`, or the default `~/.cache/huggingface/fastembed`.

## Possible Implementation Approach

- Catch embedding/storage initialization errors around REPL `_pre_load_storage()` and show a warning instead of failing the main flow.
- Guard `ContextSearchTools` initialization in `ChatAgenticNode.setup_tools()` and skip context-tool registration when unavailable.
- Add a no-op fallback for autocomplete completers so `reload_data()` failures do not affect the main REPL flow.
- Convert fastembed initialization failures into user-readable diagnostics that include network/cache details.
- Document offline deployment options: pre-cache `qdrant/all-MiniLM-L6-v2-onnx`, or configure `registry_name: openai` for embeddings.

## Acceptance Criteria

- When `huggingface.co` is unreachable and no local embedding cache exists, `datus-cli` REPL can still start and answer DB questions that do not depend on RAG.
- If `@Table`, `@Metrics`, `@Sql`, or context search is unavailable, the user sees a clear warning instead of a silent hang.
- Web and REPL degrade consistently.
- Unit tests cover storage/autocomplete/context-tools initialization failure paths.

Related: #834

stukid:
## Why

Closes #868.

In offline, intranet, or Hugging-Face-blocked environments, the default FastEmbed model may be missing from local cache and fail to download. Before this change, read-only context/autocomplete setup could initialize embeddings early and block REPL or Web context flows even when the user only needed basic chat or database tools.

## Solution

- Keep read-only vector-store operations from initializing or downloading embedding models when tables are missing or only being listed/counted.
- Convert FastEmbed initialization/download failures into clear diagnostics that name the model, Hugging Face repo, cache path, relevant environment variables, and remediation options.
- Make REPL preload, @-reference autocomplete, ChatAgenticNode context tools, Web SSE warnings, and ToolService context execution degrade at the context-search feature level while preserving DB tools and normal chat.
- Preserve explicit `limit=0` read-path semantics before dispatching to LanceDB, which treats zero as unbounded.
- Document offline deployment options for pre-caching FastEmbed artifacts or using an OpenAI-compatible embedding provider.

## Test Cases

- `uv run pytest tests/unit_tests/storage/test_base.py tests/unit_tests/storage/test_embedding_lazy_loading.py tests/unit_tests/cli/test_autocomplete.py tests/unit_tests/cli/test_repl.py tests/unit_tests/agent/node/test_chat_agentic_node.py tests/unit_tests/api/services/test_chat_task_manager.py tests/unit_tests/api/services/test_tool_service.py -q --cov=datus --cov-report=xml:/private/tmp/issue868-targeted-coverage-r2.xml --cov-report=term` -> 445 passed.
- `PATH="$PWD/.venv/bin:$PATH" diff-cover /private/tmp/issue868-targeted-coverage-r2.xml --compare-branch=upstream/main --fail-under=80` -> 90%.
- `.venv/bin/python ci/audit_tests.py --paths tests/unit_tests/storage/test_base.py tests/unit_tests/storage/test_embedding_lazy_loading.py tests/unit_tests/cli/test_autocomplete.py tests/unit_tests/cli/test_repl.py tests/unit_tests/agent/node/test_chat_agentic_node.py tests/unit_tests/api/services/test_chat_task_manager.py tests/unit_tests/api/services/test_tool_service.py` -> P0=0, P1=0.
- `.venv/bin/ruff format --check . && .venv/bin/ruff check .` -> passed.
- `PATH="$PWD/.venv/bin:$PATH" .venv/bin/python ci/run-pr-tests.py upstream/main` was attempted locally before the reviewer fix; it passed 186 acceptance tests and failed one unrelated SQLite integration test because this sandbox cannot open `/Users/kangxue/benchmark/bird/dev_20240627/dev_databases/card_games/card_games.sqlite`.

## Summary by CodeRabbit

* **New Features**
  * Graceful degradation when context-search or embedding components fail; agent, CLI and tools continue operating while surfacing a user-facing warning.

* **Improvements**
  * Startup and SSE emit a non-fatal assistant message when context-search is unavailable.
  * Autocomplete/@-reference resolution and CLI commands tolerate failures and report consolidated warnings.
  * Read-only storage queries no longer trigger embedding initialization; embedding errors provide clearer diagnostics and formatted guidance.

* **Documentation**
  * Updated storage/embedder guidance for offline/intranet setups.

* **Tests**
  * Expanded coverage for degraded-context, autocomplete, CLI, background sync, and read-only embedding behaviors.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2

### DIFF
--- a/datus/api/services/tool_service.py
+++ b/datus/api/services/tool_service.py
@@ -6,6 +6,7 @@ from typing import Any, Callable, Dict
 from datus.api.models.base_models import Result
 from datus.api.models.config_models import ErrorCode
 from datus.configuration.agent_config import AgentConfig
+from datus.storage.embedding_diagnostics import format_context_degraded_warning
 from datus.tools.func_tool.base import FuncToolResult
 from datus.tools.func_tool.context_search import ContextSearchTools
 from datus.utils.loggings import get_logger
@@ -16,9 +17,26 @@ logger = get_logger(__name__)
 class ToolService:
     """Registry-based tool dispatch: tool_name -> bound method."""
 
+    CONTEXT_TOOL_NAMES = {
+        "list_subject_tree",
+        "search_metrics",
+        "get_metrics",
+        "search_reference_sql",
+        "get_reference_sql",
+        "search_semantic_objects",
+        "search_knowledge",
+        "get_knowledge",
+    }
+
     def __init__(self, agent_config: AgentConfig):
         self._agent_config = agent_config
-        self._context_search_tools = ContextSearchTools(agent_config)
+        self.context_warning = ""
+        try:
+            self._context_search_tools = ContextSearchTools(agent_config)
+        except Exception as exc:
+            self._context_search_tools = None
+            self.context_warning = format_context_degraded_warning(exc)
+            logger.warning("Context search tool registry disabled: %s", self.context_warning)
         self._registry: Dict[str, Callable[..., FuncToolResult]] = self._build_registry()
 
     def _build_registry(self) -> Dict[str, Callable[..., FuncToolResult]]:
@@ -26,17 +44,10 @@ class ToolService:
         registry: Dict[str, Callable[..., FuncToolResult]] = {}
 
         # ContextSearchTools methods
-        context_tool_names = [
-            "list_subject_tree",
-            "search_metrics",
-            "get_metrics",
-            "search_reference_sql",
-            "get_reference_sql",
-            "search_semantic_objects",
-            "search_knowledge",
-            "get_knowledge",
-        ]
-        for name in context_tool_names:
+        if self._context_search_tools is None:
+            return registry
+
+        for name in self.CONTEXT_TOOL_NAMES:
             method = getattr(self._context_search_tools, name, None)
             if method and callable(method):
                 registry[name] = method
@@ -52,6 +63,12 @@ class ToolService:
         """Dispatch tool_name with params and return wrapped Result."""
         method = self._registry.get(tool_name)
         if method is None:
+            if self.context_warning and tool_name in self.CONTEXT_TOOL_NAMES:
+                return Result(
+                    success=False,
+                    errorCode=ErrorCode.TOOL_EXECUTION_ERROR,
+                    errorMessage=self.context_warning,
+                )
             return Result(
                 success=False,
                 errorCode=ErrorCode.TOOL_NOT_FOUND,
--- a/datus/storage/base.py
+++ b/datus/storage/base.py
@@ -178,10 +178,12 @@ class BaseEmbeddingStore(StorageBase):
         table = self._open_existing_table_for_read()
         if table is None:
             return self._empty_result(select_fields)
-        if limit:
+        if limit is not None:
             row_limit = limit
         else:
             row_limit = table.count_rows(where) if where else table.count_rows()
+        if row_limit == 0:
+            return self._empty_result(select_fields)
         result = table.search_all(where=where, select_fields=select_fields, limit=row_limit)
         if self.vector_column_name in result.column_names:
             result = result.drop([self.vector_column_name])
--- a/datus/storage/base.py
+++ b/datus/storage/base.py
@@ -136,6 +136,33 @@ class BaseEmbeddingStore(StorageBase):
                 self._create_scalar_index(col)
             logger.debug(f"Table {self.table_name} initialized successfully with embedding function")
 
+    def _open_existing_table_for_read(self) -> Optional[VectorTable]:
+        """Open an existing vector table for read-only access without embeddings."""
+        if self.table is not None:
+            return self.table
+        if not self.db.table_exists(self.table_name):
+            return None
+        self.table = self.db.open_table(self.table_name)
+        return self.table
+
+    def _empty_result(self, select_fields: Optional[List[str]] = None) -> pa.Table:
+        """Return an empty result table shaped like this store's schema."""
+        schema = self._schema or pa.schema([])
+        if select_fields is None:
+            fields = [field for field in schema if field.name != self.vector_column_name]
+        else:
+            fields = []
+            for field_name in select_fields:
+                if field_name == self.vector_column_name:
+                    continue
+                if field_name in schema.names:
+                    fields.append(schema.field(field_name))
+                else:
+                    fields.append(pa.field(field_name, pa.null()))
+        result_schema = pa.schema(fields)
+        arrays = [pa.array([], type=field.type) for field in fields]
+        return pa.Table.from_arrays(arrays, schema=result_schema)
+
     def _apply_default_values(self, data: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
         """Fill in default values for rows that are missing them."""
         if not self._default_values:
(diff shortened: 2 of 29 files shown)


datus/storage/base.py
```diff
@@ -148,12 +175,14 @@ def _apply_default_values(self, data: List[Dict[str, Any]]) -> List[Dict[str, An
...
+        table = self._open_existing_table_for_read()
+        if table is None:
+            return self._empty_result(select_fields)
         if limit:
             row_limit = limit
         else:
-            row_limit = self.table.count_rows(where) if where else self.table.count_rows()
-        result = self.table.search_all(where=where, select_fields=select_fields, limit=row_limit)
+            row_limit = table.count_rows(where) if where else table.count_rows()
+        result = table.search_all(where=where, select_fields=select_fields, limit=row_limit)
```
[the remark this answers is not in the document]
stukid: Fixed in 16392c6 by preserving explicit zero limits before backend dispatch and adding a regression test for `_search_all(limit=0)`.

datus/storage/base.py
```diff
@@ -587,10 +624,12 @@ def query_with_filter(
...
         """Query rows with filter, field selection, and optional limit."""
-        self._ensure_table_ready()
+        table = self._open_existing_table_for_read()
+        if table is None:
+            return self._empty_result(select_fields)
         if limit is None:
-            limit = self.table.count_rows(where)
-        return self.table.search_all(
+            limit = table.count_rows(where)
+        return table.search_all(
```
stukid, line 660: Please also preserve explicit `limit=0` here before calling the backend. `_search_all()` now returns an empty table for zero limits, but `query_with_filter(limit=0)` still falls through to `table.search_all(..., limit=0)`. In the current LanceDB backend, `search().limit(0)` behaves as unbounded and returns all rows, so this reintroduces the same boundary bug for callers using `query_with_filter()`. Please add an early `if limit == 0: return self._empty_result(select_fields)` and a regression test for `query_with_filter(limit=0)`.

datus/storage/base.py
```diff
@@ -148,12 +175,16 @@ def _apply_default_values(self, data: List[Dict[str, Any]]) -> List[Dict[str, An
     def _search_all(
         self, where: WhereExpr = None, select_fields: Optional[List[str]] = None, limit: Optional[int] = None
     ) -> pa.Table:
-        self._ensure_table_ready()
-        if limit:
+        table = self._open_existing_table_for_read()
```
stukid, line 178: This read-only helper only protects callers that reach it directly. Some public read/list wrappers still call `_ensure_table_ready()` before delegating, which forces embedding initialization and bypasses the new degradation behavior. In current head, `datus/storage/schema_metadata/store.py:123` (`SchemaStorage.search_all`) and `datus/storage/document/store.py:224`, `:243`, and `:382` (`list_versions`, `get_stats`, `get_all_rows`) still do this; with a failed embedding model and an existing table they raise `MODEL_EMBEDDING_ERROR` instead of reading through `open_table()`. Please remove the eager `_ensure_table_ready()` from those read-only wrappers and cover them with regression tests.

datus/api/services/tool_service.py
```diff
@@ -16,27 +17,37 @@
...
+        "search_knowledge",
+        "get_knowledge",
+    }
+
     def __init__(self, agent_config: AgentConfig):
         self._agent_config = agent_config
-        self._context_search_tools = ContextSearchTools(agent_config)
+        self.context_warning = ""
+        try:
+            self._context_search_tools = ContextSearchTools(agent_config)
```
stukid, line 35: This fail-soft path covers `ToolService`, but the CLI command path still constructs `ContextSearchTools` directly in `datus/cli/agent_commands.py:64`. `cmd_search_metrics()` and `cmd_search_reference_sql()` call that property without handling constructor failures, so an offline or missing FastEmbed cache can still crash those interactive commands instead of showing the non-fatal degraded-context warning. Please apply the same guarded construction and controlled failure there.

datus/storage/base.py
stukid: Fixed in 5c3e0bb0. Added `if limit == 0: return self._empty_result(select_fields)` early return in `query_with_filter()`, matching the existing pattern in `_search_all()`. Regression test `test_query_with_filter_zero_limit_returns_empty` confirms `limit=0` returns an empty table.

datus/storage/base.py
stukid: Fixed in 5c3e0bb0. Removed the eager `_ensure_table_ready()` from `BaseMetadataStorage.search_all()`, `DocumentStore.list_versions()`, `DocumentStore.get_stats()`, and `DocumentStore.get_all_rows()`. These all delegate to `_search_all()` which already uses `_open_existing_table_for_read()` internally, so the eager call was redundant and forced embedding initialization. Regression tests confirm these read-only wrappers return empty results without initializing embeddings on empty stores.

datus/api/services/tool_service.py
stukid: Fixed in 5c3e0bb0. Applied the same guarded construction pattern from `ToolService` to `AgentCommands.context_search_tools`: the property now catches constructor failures, stores the degraded-context warning, and returns `None`. Both `cmd_search_metrics()` and `cmd_search_reference_sql()` check for `None` and print the warning instead of crashing. Regression tests verify both commands show the degraded warning when embedding initialization fails.

PR #888, opened 2026-05-27
[BugFix] Parse legacy JSON array content in extract_user_input()

### LINKED ISSUE #887: [BugFix] Legacy session content displayed as raw JSON in Web sidebar
## Description

When viewing chat history in the Datus Web sidebar, sessions created by older versions display raw JSON strings instead of the actual user message.

**Affected format stored in DB:**
```json
[{"type": "enhanced", "content": "Database Context: \n\n**Dialect**: starrocks, **Catalog**: default_catalog, **Database**: _starrocks_audit_db_"}, {"type": "user", "content": "What can you do?"}]
```

**Expected display:** `What can you do?`  
**Actual display:** `[{"type": "enhanced", "content": "Database Context: \n\n**Dialect**:...`

## Screenshot

Legacy session JSON display

## Root Cause

`extract_user_input()` in `datus/utils/message_utils.py` handles three content shapes:
1. Python `list` of content blocks
2. `<system_reminder>...</system_reminder>\n{user}` envelope string
3. Plain string (returned unchanged)

Legacy sessions stored content as a **JSON array string** (not a Python list). Since it doesn't start with `<system_reminder>`, it falls through to case 3 and is returned as-is.

## Proposed Fix

1. Add detection for JSON array strings starting with `[` — parse with `json.loads()` and recurse into the list handler
2. Add `"user"` to recognized block types so `{"type": "user", "content": "..."}` blocks are extracted

## Environment

- Datus version: 0.3.x (upgraded from older versions that used the legacy format)
- Affected component: `datus/utils/message_utils.py` → `extract_user_input()`

FrankSen:
## Why

Sessions created by older versions stored user message content as a JSON array string:
```json
[{"type": "enhanced", "content": "Database Context: ..."}, {"type": "user", "content": "What can you do?"}]
```

`extract_user_input()` did not recognize this format and returned the raw JSON string, causing the Web sidebar history to display unreadable JSON text instead of the user's actual message.

Fixes #887

## Solution

1. Added detection for JSON array strings (starting with `[`) in `extract_user_input()` — parses with `json.loads()` and recurses into the existing list handler
2. Added `"user"` to recognized block types (`_LEGACY_USER_BLOCK_TYPES`) so `{"type": "user", "content": "..."}` blocks are properly extracted
3. Both the string-encoded and already-parsed list paths now handle legacy content

## Test Cases

- Updated existing parametrized test `test_extract_user_input_non_envelope_returns_unchanged[legacy_json_array]` to expect the extracted user text instead of raw JSON
- All 49 existing tests in `test_message_utils.py` pass
- Manually verified with real legacy session data format

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message processing to recognize additional legacy user-message formats, so user text is extracted correctly from older message structures.

* **Tests**
  * Updated unit tests to reflect the new extraction behavior for legacy JSON-array inputs.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4


stukid:
## Review: fixes the reported case correctly, but the `startswith("[")` heuristic silently blanks legitimate JSON-array messages

Nice — this does fix #887 for the actual legacy format. I verified the target case:

```
[{"type":"enhanced","content":"Database Context: ..."},{"type":"user","content":"What can you do?"}]
   -> "What can you do?"   ✅  (enhanced block correctly dropped, only the user text kept)
```

### Main concern: silent data loss on non-legacy JSON-array messages

`content.startswith("[")` is too broad. Any plain user message that *happens* to be a parseable JSON array — but isn't legacy session data — gets parsed into the list handler, where `if not isinstance(part, dict): continue` plus the type filter make it return an **empty string**. Reproduced against the patched logic:

```
'["hello", "world"]'   -> ''   ⚠️ user text lost
'[1, 2, 3]'            -> ''   ⚠️
'[{"foo":"bar"}]'      -> ''   ⚠️ (array of dicts with no recognized "type")
'[not json, just text]'-> '[not json, just text]'   ✅ (invalid JSON, falls through)
```

Before this PR those returned the raw string (ugly but readable); after, they render blank. For a history-display function, silently dropping the message is arguably worse than the original symptom. In a SQL/data tool, users pasting a JSON array as a message isn't far-fetched.

### Suggested tightening — only treat it as legacy when it actually looks like legacy blocks

```python
if content.startswith("["):
    try:
        parts = json.loads(content)
    except (json.JSONDecodeError, TypeError):
        parts = None
    if isinstance(parts, list) and any(
        isinstance(p, dict) and p.get("type") in _TEXT_BLOCK_TYPES + _LEGACY_USER_BLOCK_TYPES
        for p in parts
    ):
        return extract_user_input(parts)
# else fall through and return content unchanged
```

This keeps the #887 fix while preserving ordinary JSON-array messages.

### Test coverage is thin

The PR only flips one existing parametrized case (`legacy_json_array`) and adds no new tests for the new branches. Please add:
1. the full issue example (enhanced + user → only user extracted);
2. the already-parsed list path (`[{"type":"user","content":"..."}]`, bypassing the string branch);
3. a **regression guard** asserting that a non-legacy array message (e.g. `["hello","world"]` or `[{"foo":"bar"}]`) is returned unchanged — note the current implementation fails this (returns `""`), so it doubles as the driver for the tightening above.

### Minor / notes

- `part.get("content") or part.get("text")` drops a legitimately empty (`""`) content to the `text` fallback — edge case, low impact.
- Scope looks right: only `"user"` is treated as a legacy user block, so `enhanced`/context is correctly excluded.
- Out of scope, but worth a quick check: if the enhanced-context display path also reads this legacy format, `extract_enhanced_context()` may have a sibling gap.

### Verdict

Direction is correct and the target bug is fixed, but there's one sharp edge that silently loses data. I'd suggest, before merge: (1) tighten the legacy detection to "array contains a recognized-type block", and (2) add the tests above (especially the non-legacy-array regression guard). Happy to push those changes if helpful.

FrankSen:
Thanks for the thorough review @stukid — all three points addressed in 9cfd2cbc:

### 1. Tightened legacy detection

Adopted your suggested guard. `startswith("[")` now only recurses when the parsed array **actually contains a recognized block type**:

```python
if isinstance(parts, list) and any(
    isinstance(p, dict) and p.get("type") in _TEXT_BLOCK_TYPES + _LEGACY_USER_BLOCK_TYPES
    for p in parts
):
    return extract_user_input(parts)
```

Non-legacy JSON arrays (`'["hello","world"]'`, `'[1,2,3]'`, `'[{"foo":"bar"}]'`, `'[]'`) now fall through and return unchanged.

### 2. Tests added (9 new cases, 58 total)

- **5 regression guards** for non-legacy JSON arrays (string array, int array, dict without recognized type, unknown type, empty array) — all assert the raw string is returned unchanged
- **Full #887 scenario**: `enhanced` + `user` JSON string → only user text extracted
- **List-path legacy block**: already-parsed `[{"type":"user",...}]` list directly
- **`or`-shortcircuit edge cases**: `content=""` must NOT fallback to `text`; `content=None` must fallback

### 3. Fixed `or`-shortcircuit (minor)

Changed `part.get("content") or part.get("text")` → `part.get("content") if part.get("content") is not None else part.get("text")` so an explicitly empty `content=""` is preserved.

---

> Out of scope, but worth a quick check: if the enhanced-context display path also reads this legacy format, `extract_enhanced_context()` may have a sibling gap.

Noted — `extract_enhanced_context()` only handles the `<system_reminder>` envelope today and returns `None` for everything else, so legacy JSON arrays are already safely ignored. Will track separately if needed.

FrankSen:
CI status update for 9cfd2cbc:

✅ format-check, test-audit, title-check, CodeRabbit, sync-checklist all pass
❌ **run-coverage / coverage** failed — but it's an infrastructure flake, not a code issue:

```
ERROR: Could not install requirement datus-metricflow@ https://github.com/Datus-ai/metricflow/archive/...tar.gz
  HTTP error 504 Server Error: Gateway Time-out
```

`uv sync` failed pulling the `metricflow` tarball from GitHub (504), so the downstream coverage step errored on empty metrics (`could not convert string to float: ''`).

Could a maintainer kindly rerun the failed job? I don't have admin rights to trigger a rerun myself. The 58 unit tests pass locally (`pytest tests/unit_tests/utils/test_message_utils.py`).

PR #952, opened 2026-06-04
[BugFix] Fix run_in_terminal_sync teardown race with _is_running guard (#617)

### LINKED ISSUE #617: [Bug]: RuntimeWarning: coroutine 'run_in_terminal.<locals>.run' was never awaited on CLI exit
## Bug Description

When exiting the Datus CLI (TUI mode), Python emits an unawaited-coroutine `RuntimeWarning` originating from `prompt_toolkit.application.run_in_terminal`. The warning is harmless to functionality but indicates a coroutine is scheduled during shutdown and garbage-collected before the event loop can await it — usually a sign the TUI teardown order races with in-flight `run_in_terminal(...)` calls.

## Steps to Reproduce

1. Launch `datus` (TUI mode on a TTY)
2. Exit the session via Ctrl+C or Ctrl+D (reporter does not recall which — warning may fire on either; please verify both)
3. Observe the warning printed after the prompt returns

## Expected Behavior

Clean exit with no warnings on stderr.

## Actual Behavior

Shell prints:

```
.../python3.12/asyncio/base_events.py:711: RuntimeWarning: coroutine 'run_in_terminal.<locals>.run' was never awaited
  self._ready.clear()
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

## Environment

- OS: macOS 15.2 (Darwin 24.2.0)
- Python Version: 3.12 (conda env)
- Datus Version: 0.2.6

## Logs and Error Messages

```
<path>/lib/python3.12/asyncio/base_events.py:711: RuntimeWarning: coroutine 'run_in_terminal.<locals>.run' was never awaited
  self._ready.clear()
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

## Additional Context

Suspected source: `datus/cli/tui/console_bridge.py::run_in_terminal_sync`

```python
future = run_in_terminal(func)
...
if on_event_loop:
    return  # returns without awaiting the Future
```

`prompt_toolkit.application.run_in_terminal` internally wraps `func` in a coroutine and schedules it via `ensure_future` / `loop.create_task`. If the CLI shuts down (SIGINT / EOF) **after** a caller invokes `run_in_terminal_sync` from the event loop (`on_event_loop == True` branch → early return) but **before** prompt_toolkit's scheduler can pump the task, the wrapped coroutine is GC'd without being awaited — producing exactly this warning at `base_events.py:711` (the `_ready` queue being cleared during `loop.close()`).

Possible fixes to consider:

1. On shutdown, flush/cancel any pending `run_in_terminal` futures before closing the loop.
2. Track the returned `Future` from `run_in_terminal_sync` and `await` / cancel it during app teardown rather than discarding it when `on_event_loop` is True.
3. Suppress via `warnings.filterwarnings` only as a last resort — the right fix is deterministic cleanup.

Re-running with `PYTHONTRACEMALLOC=1` would give the allocation traceback to confirm the exact scheduling site.

nothankyouzzz:
## Why

On the event-loop branch of `run_in_terminal_sync`, calling `run_in_terminal(func)` during prompt_toolkit teardown schedules a Task whose first `__step` lands in `loop._ready`. When the loop is closed before that step runs, `loop.close()` calls `_ready.clear()`, discarding the step — leaving the inner coroutine GC'd unawaited and emitting:

```
RuntimeWarning: coroutine 'run_in_terminal.<locals>.run' was never awaited
```

Fixes #617.

## Solution

Check `app._is_running` before scheduling. prompt_toolkit sets this flag to `False` during teardown (before the loop closes), so it reliably identifies the window where scheduling would orphan a Task. When `_is_running` is `False`, call `func()` directly:

```python
if on_event_loop:
    if not getattr(app, "_is_running", True):
        # app is mid-teardown: loop.close() will call _ready.clear(),
        # discarding any newly scheduled Task's first __step.
        func()
        return
    run_in_terminal(func)
    return
```

**Why `asyncio.ensure_future()` wrapping is not sufficient** (addresses the approach in the original PR description): `run_in_terminal()` already calls `ensure_future(run())` internally and returns the resulting Task. So `ensure_future(that_task)` is a documented no-op — `ensure_future(task) is task → True`. The race is in `loop.close()` discarding `_ready`, not in whether the Task is referenced.

**Related fire-and-forget issues found during investigation** (informational only — not addressed in this PR; each requires a different strategy tied to its event loop):

- `datus/api/services/datus_service_cache.py:106,126` — untracked `asyncio.create_task()` for LRU eviction and deferred shutdown on the FastAPI loop
- `datus/api/routes/chat_routes.py:570` — untracked `asyncio.create_task(_safe_post_chat(...))` for billing/audit hook in a `finally` block

## Test Cases

Updated `tests/unit_tests/cli/tui/test_console_bridge.py`:

- `test_schedules_via_run_in_terminal_when_on_event_loop` — updated: removed the tautological `ensure_future` assertion (it asserted on a Mock, not a real Task); added `fake_app._is_running = True` to clearly exercise the live-app path
- `test_calls_directly_when_app_is_shutting_down` — **new**: sets `_is_running=False`, asserts `func()` is called directly and `run_in_terminal` is NOT called; directly exercises the teardown guard

### DIFF
--- a/datus/cli/tui/console_bridge.py
+++ b/datus/cli/tui/console_bridge.py
@@ -44,8 +44,10 @@ def run_in_terminal_sync(func: Callable[[], None]) -> None:
       avoid; ``in_terminal()`` would blank the full-screen layout for
       every print, which is destructive.
     * **On the prompt_toolkit event loop of a non-full-screen
-      Application** (e.g. legacy modals): dispatch via
-      :func:`run_in_terminal` and return immediately.
+      Application** (e.g. legacy modals): if the app is still running,
+      dispatch via :func:`run_in_terminal` and return immediately; if
+      the app is shutting down (``_is_running`` is ``False``), invoke
+      directly to avoid orphaning a Task on the closing loop.
     * **Off-loop thread with a non-full-screen Application**: submit via
       :func:`asyncio.run_coroutine_threadsafe` and block on completion.
     """
@@ -70,8 +72,15 @@ def run_in_terminal_sync(func: Callable[[], None]) -> None:
         on_event_loop = False
 
     if on_event_loop:
-        # Fire and forget — the callback is scheduled on the same loop and
-        # will run before control returns to the key handler.
+        if not getattr(app, "_is_running", True):
+            # The app is tearing down: loop.close() will call _ready.clear(),
+            # discarding any newly scheduled Task's first __step before it
+            # runs and leaving the inner coroutine GC'd unawaited.
+            func()
+            return
+        # Fire and forget — run_in_terminal schedules its own Task internally
+        # and returns it; we do not await because blocking on the loop that
+        # must run the callback would deadlock.
         run_in_terminal(func)
         return
 
--- a/tests/unit_tests/cli/tui/test_console_bridge.py
+++ b/tests/unit_tests/cli/tui/test_console_bridge.py
@@ -35,29 +35,55 @@ def test_runs_inline_when_no_application() -> None:
 
 
 def test_schedules_via_run_in_terminal_when_on_event_loop() -> None:
-    """On the prompt_toolkit event loop, dispatch fires-and-forgets.
+    """On the prompt_toolkit event loop with a live app, dispatch via run_in_terminal.
 
-    Fire-and-forget because blocking on the returned future would deadlock
-    the loop that has to run the scheduled callback. The no-await path is
-    taken only when ``asyncio.get_running_loop()`` succeeds on the calling
-    thread, so we simulate that with a patch.
+    Fire-and-forget because blocking on the returned awaitable would deadlock
+    the loop that has to run the scheduled callback. run_in_terminal already
+    schedules its own Task internally (ensure_future(run())) and returns it,
+    so no outer wrapping is needed.
     """
-    fake_future = mock.MagicMock()
     fake_app = mock.MagicMock()
     fake_app.full_screen = False  # legacy modal path; full_screen path is exercised separately
+    fake_app._is_running = True
     fake_loop = mock.MagicMock()
 
     with (
         mock.patch("datus.cli.tui.console_bridge.get_app_or_none", return_value=fake_app),
         mock.patch("datus.cli.tui.console_bridge.asyncio.get_running_loop", return_value=fake_loop),
-        mock.patch("datus.cli.tui.console_bridge.run_in_terminal", return_value=fake_future) as mocked_rit,
+        mock.patch("datus.cli.tui.console_bridge.run_in_terminal") as mocked_rit,
     ):
         fn = mock.MagicMock()
         run_in_terminal_sync(fn)
 
         mocked_rit.assert_called_once_with(fn)
-        # On-loop path must NOT block on the future — it would deadlock.
-        fake_future.result.assert_not_called()
+
+
+def test_calls_directly_when_app_is_shutting_down() -> None:
+    """On the event loop but with _is_running=False (app teardown), func must
+    be called inline rather than scheduled.
+
+    Root cause of #617: loop.close() calls _ready.clear(), discarding the
+    Task's pending first __step before it runs. The inner coroutine is then
+    GC'd unawaited, emitting RuntimeWarning. Detecting _is_running=False and
+    falling back to a direct call eliminates the race.
+    """
+    fake_app = mock.MagicMock()
+    fake_app.full_screen = False
+    fake_app._is_running = False
+    called = {"count": 0}
+
+    def _fn() -> None:
+        called["count"] += 1
+
+    with (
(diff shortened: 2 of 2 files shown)


stukid:
## Review: this change is effectively a runtime no-op and likely does not fix #617

@nothankyouzzz  Thanks for digging into this. I went through the prompt_toolkit internals and reproduced the warning deterministically — unfortunately I don't think wrapping in `asyncio.ensure_future` addresses the root cause.

### Why it's a no-op at runtime

The PR description assumes `run_in_terminal(func)` returns an *unscheduled* awaitable that gets orphaned. It doesn't. In prompt_toolkit 3.0.52, `run_in_terminal` already schedules its inner coroutine itself:

```python
# prompt_toolkit/application/run_in_terminal.py
def run_in_terminal(func, ...) -> Awaitable[_T]:
    async def run(): ...
    return ensure_future(run())   # ensure_future IS asyncio.ensure_future here
```

So `run_in_terminal(func)` returns an already-scheduled `Task`. And `asyncio.ensure_future(<a Task>)` returns that **same Task unchanged** (documented behavior — verified locally, `ensure_future(task) is task` → `True`). The inner `run()` coroutine is wrapped into the Task by prompt_toolkit at the exact same moment in both old and new code; the outer wrap never touches `run()` and the return value is discarded just like before. Net runtime behavior is identical.

### The actual root cause (matches the analysis in #617)

The warning (`base_events.py:711`, `self._ready.clear()`) is a teardown race: `run_in_terminal_sync` is called on the event-loop branch, prompt_toolkit schedules the task's first `__step` into `_ready`, then the loop is closed (Ctrl-C / EOF) before that step is pumped → the inner coroutine is GC'd unawaited. Re-wrapping the already-built Task changes nothing about this.

### Deterministic reproduction (old vs this PR vs a real fix)

I reproduced the exact race in a script (force-schedule, then `loop.stop()` before the step runs, then `close()` + `gc.collect()` under `warnings.catch_warnings`):

```
OLD code  run_in_terminal(func)             -> WARNING reproduced
PR #952   ensure_future(run_in_terminal())  -> WARNING reproduced   <-- same as old
REAL FIX  hold ref + cancel/drain at exit   -> clean (no warning)
```

Both old and new emit `coroutine 'run_in_terminal.<locals>.run' was never awaited`.

### The updated test gives false confidence

The test mocks `run_in_terminal` to return a `MagicMock` and asserts `asyncio.ensure_future` was called with it. That passes, but in reality `run_in_terminal` returns a real `Task`, so `ensure_future` is a no-op — the assertion is essentially tautological and never exercises the warning. A meaningful regression test should drive the near-shutdown path under `warnings.catch_warnings(record=True)` and assert **no** `RuntimeWarning`; this PR would not pass that test.

### Suggested direction (from #617)

Deterministic cleanup rather than re-wrapping, e.g. one of:
1. Track the futures returned on the on-loop branch and `cancel()`/`await` them during TUI teardown before the loop closes; or
2. When the app/loop is already shutting down, fall back to calling `func()` directly instead of scheduling.

Plus a regression test that actually captures `RuntimeWarning` on that path.

### Caveat

I verified the mechanism via source + a deterministic asyncio repro, not a live TTY-exit run (no interactive terminal here), so I'd suggest confirming with `PYTHONWARNINGS=always::RuntimeWarning PYTHONTRACEMALLOC=1 datus` and reproducing the exit race on both `main` and this branch — I expect the warning to persist on both.

Happy to put up an alternative PR (drain/cancel at teardown + a warning-capturing regression test) if useful.

nothankyouzzz:
@stukid Thanks for the thorough analysis — you’re completely right. I went back through the prompt_toolkit source and confirmed `ensure_future(task) is task → True`; the original fix was a no-op at runtime and the updated test was asserting on a Mock, not a real Task, so it gave false confidence.

This PR is now updated with the `_is_running` guard approach (option 2 from your comment):

```python
if on_event_loop:
    if not getattr(app, "_is_running", True):
        # app is mid-teardown: loop.close() calls _ready.clear(),
        # discarding any newly scheduled Task's first __step.
        func()
        return
    run_in_terminal(func)
    return
```

Changes:
- **`console_bridge.py`**: checks `app._is_running` before calling `run_in_terminal`. When `False` (prompt_toolkit sets this before the loop closes), falls back to a direct `func()` call.
- **`test_console_bridge.py`**: added `test_calls_directly_when_app_is_shutting_down` — sets `_is_running=False`, asserts `func()` is called and `run_in_terminal` is NOT called. Also fixed the existing live-app test to set `_is_running=True` explicitly.

The race was verified using the same `call_soon + loop.stop()` + `warnings.catch_warnings(record=True)` pattern you described — the raw `run_in_terminal()` call always emits the warning; the `run_in_terminal_sync()` path is clean on this branch.

The branch has been rebased onto current main.

PR #964, opened 2026-06-05
[BugFix] Migrate stale UNIQUE constraint on subject_nodes for existing DBs (#507)

nothankyouzzz:
## Why

PR #493 (Storage 2.0) changed the `subject_nodes` unique constraint from `UNIQUE(parent_id, name)` to `UNIQUE(parent_id, name, datasource_id)` to allow the same node name to exist under different datasources. However, SQLite does not support `ALTER TABLE DROP CONSTRAINT`, so existing databases retained the old constraint even after upgrading. The existing `_migrate_missing_columns` migration correctly added the `datasource_id` column but had no way to fix the stale constraint. As a result, any cross-datasource insert with a duplicate `(parent_id, name)` pair would fail with `UNIQUE constraint failed` on upgraded databases.

Fixes #507.

## Solution

Add `_migrate_constraints()` to `SqliteRdbDatabase`, called from `ensure_table()` after `_migrate_missing_columns()`. The method:

1. Reads the table's original `CREATE TABLE` statement from `sqlite_master`
2. Extracts and normalizes all `UNIQUE(...)` constraints from it
3. Compares against the desired constraints in `TableDefinition`
4. If they differ, performs a full table rebuild: `CREATE TABLE tmp → INSERT SELECT → DROP old → RENAME`

The rebuild is atomic (runs inside the same transaction as the rest of `ensure_table`) and is a no-op on fresh installs where the constraint is already correct. `_migrate_missing_columns` runs first so newly added columns are included in the data copy during rebuild.

## Test Cases

Added `TestMigrateConstraints` to `tests/unit_tests/storage/rdb/test_sqlite_backend.py` with 4 unit tests:

- `test_noop_when_table_missing` — no rebuild attempted on an empty database
- `test_noop_when_constraints_match` — no rebuild when constraint is already correct; existing data survives
- `test_rebuilds_stale_constraint_and_preserves_data` — full upgrade path: old DB is rebuilt, data is preserved, cross-datasource inserts now succeed
- `test_old_constraint_still_blocks_same_datasource` — new constraint still rejects true duplicates after rebuild

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database table constraint handling to properly rebuild tables when constraint requirements change while preserving existing data.

* **Tests**
  * Added test coverage for constraint migration scenarios to ensure data integrity during table rebuilds.

### DIFF
--- a/datus/storage/rdb/sqlite_backend.py
+++ b/datus/storage/rdb/sqlite_backend.py
@@ -233,6 +233,10 @@ class SqliteRdbDatabase(RdbDatabase):
             with self._auto_conn() as conn:
                 # Migrate missing columns on existing tables before running DDL
                 self._migrate_missing_columns(conn, table_def)
+                # Rebuild table if inline constraints differ from the definition.
+                # Must run after _migrate_missing_columns so newly added columns
+                # are included in the data copy.
+                self._migrate_constraints(conn, table_def)
                 for stmt in ddl_statements:
                     conn.execute(stmt)
         except DatusException:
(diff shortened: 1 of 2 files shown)


stukid:
Thanks for contributing! I'll merge it.

PR #1042, opened 2026-06-18
[BugFix] Warn on unknown --subagent names instead of silent gen_sql fallback

### LINKED ISSUE #1009: [Bug]: --subagent silently falls back to a gen_sql node for unknown subagent names
## Bug Description

`datus -p ... --subagent <name>` performs no validation of the subagent name. Any unrecognized name — a typo, a stale name like `generate_sql` (renamed to `gen_sql` in 0.3.4, #935), or a completely bogus string — is silently routed to the catch-all branch in the node factory and becomes a `GenSQLAgenticNode`. The run exits 0 and answers normally, so users have no way to tell that their intended (custom) subagent was never loaded. This is especially confusing right after the 0.3.4 `gen_sql` rename: projects still passing `generate_sql` get no migration signal at all.

## Steps to Reproduce

1. In any configured project, run: `datus -p 'hi' --subagent totally_bogus_name_xyz --stream`
2. Observe the run completes with exit code 0 and streams a normal answer (a SQL-assistant persona), with no error or warning on stdout/stderr.
3. Same behavior with the pre-0.3.4 name: `datus -p 'hi' --subagent generate_sql --stream`

## Expected Behavior

An unknown subagent name should fail fast with an error (or at minimum print a warning) listing the available subagent names, instead of silently running a different agent.

## Actual Behavior

The CLI silently instantiates `GenSQLAgenticNode` named after the unknown string and answers as if nothing was wrong (exit 0, no warning).

## Environment

- OS: macOS (Darwin 24.2.0)
- Python Version: 3.12.8
- Datus Version: 0.3.4rc1 (PyPI wheel)
- Commit: `3799386a` (2026-06-11) — source checkout contains identical `node_factory.py` code

## Logs and Error Messages

```
$ datus -p 'hi' --subagent totally_bogus_name_xyz --stream
{"message_id":"thinking_stream_<redacted>","role":"assistant","content":[{"type":"thinking-delta", ...}]}
... normal streamed answer: "Hi! I'm your SQL assistant connected to the <db> DuckDB database..."
$ echo $?
0
(No warning or error anywhere in stdout/stderr.)
```

### Root cause

`datus/agent/node/node_factory.py:267` — `create_interactive_node` dispatches known subagent names through `if/elif` branches, then ends in a catch-all `else`:

```python
else:
    from datus.agent.node.gen_sql_agentic_node import GenSQLAgenticNode

    return GenSQLAgenticNode(
        node_id=node_id if node_id is not None else f"{subagent_name}{node_id_suffix}",
        description=f"SQL generation node for {subagent_name}",
        node_type="gen_sql",
        ...
        node_name=subagent_name,
        ...
    )
```

**Confirmed** (reproduced in this session): any string not matched by an earlier branch and not present in `agent_config.agentic_nodes` (`_resolve_node_class_type`, `node_factory.py:492`) lands here. `datus/cli/print_mode.py:50` reads `args.subagent` and passes it through (`:84`) without validation.

### Downstream impact

- Custom subagents with a typo'd name silently run as plain `gen_sql`, with the custom system prompt / tool whitelist silently dropped.
- Projects still using the pre-0.3.4 `generate_sql` name (renamed in #935) get no migration warning — behavior may drift without any signal.
- The same factory serves both REPL and print mode, so the issue affects both entry points.

### Suggested fixes

**Minimal** (lowest risk / fastest to land):
- File: `datus/agent/node/node_factory.py:267`
- Change: in the catch-all branch, if `subagent_name` is not a key in `agent_config.agentic_nodes`, log a prominent warning (`unknown subagent 'X', falling back to gen_sql; available: ...`) before constructing the node.
- Tradeoff: behavior still silently works; scripted callers may not see the warning.

**Preferred** (correct fix, more work):
- File(s): `datus/agent/node/node_factory.py`, `datus/cli/print_mode.py`
- Change: validate the name against built-in node types + configured `agentic_nodes` keys and raise a `DatusException` listing valid names; reserve the `gen_sql` fallback for names explicitly configured with `node_class`/`type: gen_sql`.
- Tradeoff: breaking for anyone (accidentally) relying on the fallback; needs a release-note entry.

### Related

- #935 (gen_sql rename) — old `generate_sql` name now hits this fallback silently.

mvanhorn:
## Why

Closes #1009.

`datus -p ... --subagent <name>` did no validation of the subagent name. Any
unrecognized name (a typo, the stale pre-0.3.4 name `generate_sql`, or a bogus
string) silently fell through the catch-all `else` branch in
`create_interactive_node` (`datus/agent/node/node_factory.py`) and became a
`GenSQLAgenticNode`. The run exited 0 and answered normally, so users got no
signal that their intended custom subagent was never loaded; its system prompt
and tool whitelist were silently dropped.

## Solution

Implements the issue's lower-risk "Minimal" variant (warn, do not raise):

- Before constructing the fallback `GenSQLAgenticNode`, detect whether
  `subagent_name` is genuinely unrecognized. It is recognized if it is a
  built-in subagent name, a key in `agent_config.agentic_nodes`, or its
  resolved `node_class_type` maps to a built-in class.
- For a genuinely unknown name, emit a `logging.WARNING` naming the unknown
  subagent and listing the available names (sorted union of built-ins and
  configured `agentic_nodes` keys), e.g.
  `Unknown subagent 'generate_sql'; falling back to gen_sql. Available subagents: ...`.
- The `gen_sql` fallback is preserved so scripted callers do not break.

A module-level `_BUILTIN_SUBAGENT_NAMES` frozenset is the single source of truth
shared by the recognition check and the warning's "available" list. The fix
lives in the factory, so every caller of `create_interactive_node` (print mode,
REPL, the chat task manager) benefits without duplicated logic.

## Test Cases

Added unit coverage in `tests/unit_tests/agent/node/test_node_factory.py`
(asserted via `caplog`):

- A built-in name (`gen_sql`) constructs its node with no warning.
- A configured `agentic_nodes` key constructs with no warning.
- An unknown name (`totally_bogus_name_xyz`) still constructs the gen_sql
  fallback AND emits a warning listing available names.
- The stale pre-0.3.4 name `generate_sql` triggers the warning (surfaces the
  #935 rename regression as a migration signal).
- The warning text includes both built-in names and configured `agentic_nodes`
  keys so the message is actionable.

`python -m pytest tests/unit_tests/agent/node/test_node_factory.py -q` -> 49 passed.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved interactive node creation when an unknown subagent is requested by falling back more safely and logging a clear warning.
  * The warning now includes the unknown name and a list of available subagents, making it easier to choose a valid option.
  * Refined which subagents appear as available, so hidden or incompatible options are no longer shown in the interactive flow.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved interactive node creation when an unknown subagent is requested by falling back more safely.
  * Added clearer warning logs that include the unknown name and a list of valid alternatives.
  * Refined which subagents are considered available so hidden or incompatible options are no longer presented during interactive flow.

### DIFF
--- a/datus/agent/node/node_factory.py
+++ b/datus/agent/node/node_factory.py
@@ -8,10 +8,43 @@ Shared factory functions for creating interactive agentic nodes and their inputs
 Used by CLI print mode and interactive REPL to avoid duplicating node creation logic.
 """
 
+import logging
 from typing import TYPE_CHECKING, Literal, Optional
 
 if TYPE_CHECKING:
     from datus.configuration.agent_config import AgentConfig
+
+
+logger = logging.getLogger(__name__)
+
+_BUILTIN_SUBAGENT_NAMES = frozenset(
+    {
+        "ask_dashboard",
+        "ask_metrics",
+        "ask_report",
+        "explore",
+        "feedback",
+        "gen_dashboard",
+        "gen_job",
+        "gen_metrics",
+        "gen_report",
+        "gen_semantic_model",
+        "gen_skill",
+        "gen_sql",
+        "gen_sql_summary",
+        "gen_table",
+        "gen_visual_dashboard",
+        "gen_visual_report",
+        "scheduler",
+    }
+)
+
+
+def _configured_agentic_node_names(agent_config: "AgentConfig") -> frozenset[str]:
+    agentic_nodes = getattr(agent_config, "agentic_nodes", None)
+    if not agentic_nodes or not hasattr(agentic_nodes, "keys"):
+        return frozenset()
+    return frozenset(str(name) for name in agentic_nodes.keys())
 
 
 def create_interactive_node(
@@ -265,6 +298,19 @@ def create_interactive_node(
             )
 
         else:
+            configured_agentic_node_names = _configured_agentic_node_names(agent_config)
+            if (
+                subagent_name not in _BUILTIN_SUBAGENT_NAMES
+                and subagent_name not in configured_agentic_node_names
+                and node_class_type not in _BUILTIN_SUBAGENT_NAMES
+            ):
+                available_names = ", ".join(sorted(_BUILTIN_SUBAGENT_NAMES | configured_agentic_node_names))
+                logger.warning(
+                    "Unknown subagent %r; falling back to gen_sql. Available subagents: %s",
+                    subagent_name,
+                    available_names,
+                )
+
             from datus.agent.node.gen_sql_agentic_node import GenSQLAgenticNode
 
             return GenSQLAgenticNode(
--- a/datus/agent/node/node_factory.py
+++ b/datus/agent/node/node_factory.py
@@ -8,36 +8,16 @@ Shared factory functions for creating interactive agentic nodes and their inputs
 Used by CLI print mode and interactive REPL to avoid duplicating node creation logic.
 """
 
-import logging
 from typing import TYPE_CHECKING, Literal, Optional
+
+from datus.utils.constants import HIDDEN_SYS_SUB_AGENTS, SYS_SUB_AGENTS
+from datus.utils.loggings import get_logger
 
 if TYPE_CHECKING:
     from datus.configuration.agent_config import AgentConfig
 
 
-logger = logging.getLogger(__name__)
-
-_BUILTIN_SUBAGENT_NAMES = frozenset(
-    {
-        "ask_dashboard",
-        "ask_metrics",
-        "ask_report",
-        "explore",
-        "feedback",
-        "gen_dashboard",
-        "gen_job",
-        "gen_metrics",
-        "gen_report",
-        "gen_semantic_model",
-        "gen_skill",
-        "gen_sql",
-        "gen_sql_summary",
-        "gen_table",
-        "gen_visual_dashboard",
-        "gen_visual_report",
-        "scheduler",
-    }
-)
+logger = get_logger(__name__)
 
 
 def _configured_agentic_node_names(agent_config: "AgentConfig") -> frozenset[str]:
@@ -45,6 +25,31 @@ def _configured_agentic_node_names(agent_config: "AgentConfig") -> frozenset[str
     if not agentic_nodes or not hasattr(agentic_nodes, "keys"):
         return frozenset()
     return frozenset(str(name) for name in agentic_nodes.keys())
+
+
+def _visible_subagent_names(agent_config: "AgentConfig") -> frozenset[str]:
+    visible_names = set(SYS_SUB_AGENTS - HIDDEN_SYS_SUB_AGENTS)
+    agentic_nodes = getattr(agent_config, "agentic_nodes", None)
+    if not agentic_nodes or not hasattr(agentic_nodes, "items"):
+        return frozenset(visible_names)
+
+    current_db = getattr(agent_config, "current_datasource", None)
+    for name, sub_config in agentic_nodes.items():
+        name = str(name)
+        if name in SYS_SUB_AGENTS or name == "chat":
+            continue
+        if hasattr(sub_config, "model_dump"):
+            sub_config = sub_config.model_dump()
+        if isinstance(sub_config, dict):
+            scoped_context = sub_config.get("scoped_context") or {}
+            if hasattr(scoped_context, "model_dump"):
+                scoped_context = scoped_context.model_dump()
+            if isinstance(scoped_context, dict):
+                scoped_datasource = scoped_context.get("datasource")
+                if scoped_datasource and scoped_datasource != current_db:
+                    continue
+        visible_names.add(name)
+    return frozenset(visible_names)
 
 
 def create_interactive_node(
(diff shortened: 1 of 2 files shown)


stukid (commented):
Thanks for the fix — the core behavior is correct and well-tested, and placing the check in the factory is the right call (print mode calls `create_interactive_node` with no upstream validation, so all of print/REPL/chat benefit). A few quality points left as inline comments; none are blockers, but the source-of-truth duplication is worth addressing before merge.

Summary of inline notes:
1. (medium) `_BUILTIN_SUBAGENT_NAMES` duplicates the existing canonical `SYS_SUB_AGENTS` and the two already disagree.
2. (medium) The "Available subagents" list advertises hidden/internal agents users shouldn't be told to use.
3. (low) Logger convention — use `datus.utils.loggings.get_logger`.
4. (low) The third `if` condition is effectively redundant.

datus/agent/node/node_factory.py
```diff
@@ -8,12 +8,45 @@
...
 """
 
+import logging
 from typing import TYPE_CHECKING, Literal, Optional
 
 if TYPE_CHECKING:
     from datus.configuration.agent_config import AgentConfig
 
 
+logger = logging.getLogger(__name__)
```
stukid: **[low / convention]** Per CLAUDE.md ("Logging: `from datus.utils.loggings import get_logger` — never `print()`") and every other node in this directory (`ask_metrics_agentic_node.py`, `agentic_node.py`, `chat_agentic_node.py`, …), please use the project helper instead of the stdlib logger:

```python
from datus.utils.loggings import get_logger

logger = get_logger(__name__)
```

The logger name stays `__name__` (`datus.agent.node.node_factory`), so the `caplog`-based tests are unaffected.

datus/agent/node/node_factory.py
```diff
@@ -8,12 +8,45 @@
...
+
+_BUILTIN_SUBAGENT_NAMES = frozenset(
```
stukid: **[medium / duplicate source of truth]** There is already a canonical built-in subagent set: `SYS_SUB_AGENTS` in `datus/utils/constants.py`, imported by `repl.py`, `api/constants.py`, `autocomplete.py`, `agent_app.py`, `sub_agent_wizard.py`, and `bootstrap_bi_commands.py`. This new `_BUILTIN_SUBAGENT_NAMES` is a third hand-maintained copy, and it already **disagrees** with the canonical one:

- `SYS_SUB_AGENTS` (14): does **not** contain `explore`, `ask_report`, `ask_dashboard`
- this frozenset (17): **adds** those three

The PR description's "single source of truth" only holds within this file. Future subagent additions now have to update yet another place, and the lists are already out of sync. Suggest importing `SYS_SUB_AGENTS` and reconciling the three extra names rather than introducing a parallel list.

datus/agent/node/node_factory.py
```diff
@@ -265,6 +298,19 @@ def create_interactive_node(
             )
 
         else:
+            configured_agentic_node_names = _configured_agentic_node_names(agent_config)
+            if (
+                subagent_name not in _BUILTIN_SUBAGENT_NAMES
+                and subagent_name not in configured_agentic_node_names
+                and node_class_type not in _BUILTIN_SUBAGENT_NAMES
```
stukid: **[low / redundant condition]** `node_class_type not in _BUILTIN_SUBAGENT_NAMES` can never be the deciding factor here. `_resolve_node_class_type` returns a non-`None` value only when `subagent_name` is a key in `agent_config.agentic_nodes` — i.e. when `subagent_name in configured_agentic_node_names`, which the line above already short-circuits on. When `subagent_name` is not a configured key, `node_class_type` is `None` and `None not in _BUILTIN_SUBAGENT_NAMES` is always `True`. Harmless, but it can be dropped to simplify the check.

datus/agent/node/node_factory.py
```diff
@@ -265,6 +298,19 @@ def create_interactive_node(
...
+            ):
+                available_names = ", ".join(sorted(_BUILTIN_SUBAGENT_NAMES | configured_agentic_node_names))
```
stukid: **[medium / misleading message]** The "Available subagents" list will advertise names the rest of the system deliberately keeps out of user-facing listings:

- `feedback` is in `HIDDEN_SYS_SUB_AGENTS` — hidden from the `/agent` selector and autocomplete on purpose.
- `explore`, `ask_report`, `ask_dashboard` are not in the user-facing `SYS_SUB_AGENTS` — `explore` is an agent-to-agent delegation target (`BUILTIN_SUBAGENT_DESCRIPTIONS`) and `ask_*` are contextual follow-up nodes, not top-level `--subagent` choices.

So after a typo the warning may suggest internal/hidden agents. The canonical user-facing set is `SYS_SUB_AGENTS - HIDDEN_SYS_SUB_AGENTS` (cf. `BUILTIN_SUBAGENTS` in `api/constants.py:17` and `agent_app.py:260`). Recommend deriving `available_names` from `(SYS_SUB_AGENTS - HIDDEN_SYS_SUB_AGENTS) | configured_agentic_node_names` — this also resolves the duplicate-list concern above.

stukid:
Hi @mvanhorn, thanks for contributing! Please check these review comments.

mvanhorn:
All four addressed in b9a5e7ad:

- Switched to the project's get_logger helper per CLAUDE.md.
- Dropped the duplicate builtin set and imported the canonical SYS_SUB_AGENTS from datus.utils.constants.
- Removed the redundant node_class_type condition - you were right that it could never be the deciding factor.
- The "Available subagents" listing now excludes HIDDEN_SYS_SUB_AGENTS and datasource-scoped nodes that don't match the current datasource, mirroring how the /agent selector filters.

Thanks for the careful pointers to the existing conventions.

mvanhorn:
Added focused unit tests covering the new subagent-validation helpers — `_configured_agentic_node_names` (key normalization + non-mapping handling), `_visible_subagent_names` (hidden and datasource-scoped filtering), and the unknown-subagent warning contents — to bring diff coverage over the 80% gate.

I couldn't run the suite locally (missing `pyarrow`/`datus_db_core`), so I'm relying on CI to confirm — will follow up if anything's red.

PR #1106, opened 2026-07-07
[Refactor] Restructure the plugin system

liuyufei9527:
Replace the scheduler-specific extension subsystem with a general datus-plugin framework, then build the full plugin surface on top of it: CLI dispatch, bundled skills, system-prompt injection, namespace-scoped bash permissions, and tool argument transformers.

Framework
- New `datus.plugins` entry-point group: one entry per plugin resolving to a class datus drives by duck-typed contract (`run_cli`, optional class-level `skills_dir` / `system_prompt` / `cli_permissions` / `tool_transformers`). Plugins never import `datus.*` and depend on no shared SDK.
- Datus acts as the config broker: parses `agent.plugins.<name>.<profile>`, expands `${VAR}`, resolves the active profile (--profile > project pin in .datus/config.yml > default:true > sole profile > {}), constructs the plugin with a plain dict, and dispatches `datus <plugin> ...` in-process before argparse.
- `agent.plugins_enabled: false` master switch disables all plugin functionality (dispatch, skills, prompt injection, permissions, transformers). A datus-owned `## Plugins` preamble names the loaded config file so plugins never hard-code config paths; unconfigured plugins receive {} and point at a bundled `<name>-setup` skill.

Permissions
- Optional class-level `cli_permissions()` declares allow/ask/deny rules per permission profile (normal/auto); every pattern is prefixed with `datus <entry-point-name> ` so a plugin can never shape permissions outside its own namespace. Evaluation keeps deny > safety > ask > allow, so user rules always beat plugin rules.
- Plugin-declared ask rules can be relaxed per project: the confirmation prompt offers "allow (project)", persisting the exact matched pattern to `.datus/config.yml` `bash_allow`. Leading `--profile` globals are normalized before matching; `--config` deliberately is not.

Tool transformers
- Optional class-level `tool_transformers()` maps tool patterns (proxy syntax) to callables that rewrite or deny the agent's tool-call arguments before execution (e.g. tenant-scoping execute_sql). Implemented in datus/tools/middleware/ by rebuilding FunctionTool; sync or async, fail-closed on raise or non-dict return, applied lazily per agentic node and gated by plugins_enabled.

Removals
- SchedulerAgenticNode / SchedulerTools / scheduler-validation subsystem, `agent.services.schedulers` config, and the `/services scheduler` TUI — scheduler access is now fully owned by plugins. All scheduler-adapter references purged; docs use a fictional hello plugin.
- Vestigial `allowed_commands` skill frontmatter field (unenforced since SkillBashTool was dissolved into the generic BashTool).

Docs
- New top-level Plugin section (en/zh): plugin/introduction.md (user-facing install/config/permissions) and plugin/development.md (contract, quickstart, recipes, skills, prompt injection, permissions, transformers, setup-skill convention, testing, checklist).

Test Cases: unit tests for plugin registry (discovery, profile resolution, permission collection, transformer collection, malformed declarations), CLI dispatch, permission layering and project grants, tool middleware (rewrite/deny/async/ordering/JSON edge cases), and agentic-node application (gating, idempotence, fail-closed apply); scheduler-subsystem tests removed with the code.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
## Summary by CodeRabbit

* **New Features**
  * Added plugin support for CLI commands, skills, prompt context, permission rules, and tool argument transforms.
  * Updated configuration and service management to use plugin-based scheduler capabilities.

* **Bug Fixes**
  * Simplified services, subagents, validation, and tooling to remove scheduler-specific behavior and align menus, prompts, and outputs with the new structure.
  * Improved permission handling for plugin-based commands, including profile-aware bash rules and project-level allowances.

* **Documentation**
  * Refreshed setup, configuration, quickstart, and subagent docs to reflect the new plugin and semantic-layer workflow.

## Summary by CodeRabbit

* **New Features**
  * Added a plugin framework for CLI commands, skills, system-prompt context injection, bash permission rules, and tool-argument transformers.
  * Updated services and runtime to rely on plugin-provided “scheduler” capability rather than built-in scheduler handling.
* **Bug Fixes**
  * Tightened permission behavior with profile-aware plugin rules and project-level allowances.
  * Improved tool handling to apply plugin transformers safely and consistently.
* **Documentation**
  * Updated configuration, quickstart, and service/TUI docs to remove scheduler UI/docs and focus on Dashboard + Semantic Layer, with new plugin guidance.


liuyufei9527:
Thanks for the review — addressed in the latest push. Summary of what changed and what was intentionally left as-is.

### Fixed
- **`is_enabled` preserved on tool rebuild** — `wrap_tool_with_transformers` now carries over `is_enabled` (plus both guardrail fields) so a disabled/gated tool can't be silently re-enabled when it matches a transformer pattern. Note: this SDK version has no `custom_data_extractor` field; the equivalents are `tool_input_guardrails` / `tool_output_guardrails`, which are now carried over. Added `test_wrapped_tool_preserves_disabled_state`.
- **Public pattern API** — `proxy_tool` now exposes `parse_tool_patterns` / `tool_name_matches`; the middleware imports those instead of the underscore-prefixed internals (private aliases kept for existing call sites/tests).
- **Docs** — dropped the stale scheduler wording on the datasources pages (en/zh); fixed the Chinese plugin docs to link to `.zh.md` counterparts; corrected the subagent count (thirteen → twelve / 十三 → 十二); added a Plugins entry to the configuration Next Steps (en/zh) and removed another lingering "调度器" reference in the zh intro.
- **Stale docstring** — `_Entry` now points at `_reload_entries` (`_build_dashboard_entries` / `_build_semantic_entries`).
- Coverage-map exclude path and the two new plugin nav pages were already handled in the prior push.

### Already resolved in the current code (flagged against an earlier revision)
- **Fail-closed transformer flag** — `_ensure_tool_transformers` only sets `_tool_transformers_applied` after a successful wrap (and on the no-op paths); the exception path leaves it unset and re-raises, so retries stay loud. `apply_tool_transformers` builds the new tool list and only then writes `tools[:]`, so a mid-failure never leaves partial wrapping (retry-safe, no double-wrap).
- **CLI dispatch crash-safety** — both the adapter `handler(...)` call and `plugin.run_cli(...)` are wrapped in try/except, and return codes go through `_coerce_exit_code`, which maps a non-int return to a logged warning + exit 0 rather than raising.
- **Exact plugin ask-rule project grants** — an ask-rule hit persists the exact matched pattern (never widened to `:*`), which is the same key the strict-equality grant bypass checks, so exact plugin ask rules round-trip correctly.

### Intentionally skipped
- **`main.py` — use `print_error` instead of `print(..., file=sys.stderr)`**: plugin/adapter dispatch runs before the Rich CLI UI is initialized; keeping these on plain stderr is deliberate (scriptable, no ANSI), and `main.py` doesn't otherwise depend on `cli_styles`.
- **Rename `test_plugin_dispatch.py` into `test_main.py`**: `main.py`'s tests are intentionally split by concern (`test_main` / `test_external_command_dispatch` / `test_plugin_dispatch`); merging adds file size with no coverage gain, and the test-quality audit doesn't flag it.

PR #1127, opened 2026-07-10
[BugFix] Fix --task_db_name parameter being ignored in current_db_name_type

### LINKED ISSUE #1126: [BugFix] --task_db_name parameter is ignored, datasource name used instead
## Bug Description

When running `datus-agent run --datasource mysql_test --task "..." --task_db_name fin_risk_ctrl`, the `--task_db_name` value is ignored and the datasource name (`mysql_test`) is used as the target database name.

## Expected Behavior

The database specified by `--task_db_name` should be used for the task.

## Actual Behavior

From the logs:
```
database_name='mysql_test'  # should be 'fin_risk_ctrl'
```

Then schema linking fails with:
```
(1049, "Unknown database 'mysql_test'")
```

## Root Cause

In `datus/configuration/agent_config.py:1797`, the `current_db_name_type()` method has a fallback path that returns `self._current_datasource` when the `db_name` argument (from `--task_db_name`) is not a key in the datasources dictionary. Since the user's database name is typically a real database name (not a datasource config key), it never matches and gets discarded.

## Fix

Change line 1797 from:
```python
return self._current_datasource, datasources[self._current_datasource].type
```
to:
```python
return db_name or self._current_datasource, datasources[self._current_datasource].type
```

PR: https://github.com/Datus-ai/Datus-agent/pull/new/fix/task-db-name-parameter

pika-Ysw:
## Why

When running `datus-agent run --datasource X --task "..." --task_db_name Y`, the `--task_db_name` value is ignored and the datasource name is used as the database name instead. The log showed `database_name='X'` when it should have been `database_name='Y'`. Closes #1126.

## Solution

The `current_db_name_type()` method in `agent_config.py` has a fallback path (line 1797) that returns `self._current_datasource` when `db_name` (from `--task_db_name`) is not a datasource key. This discards the user-provided database name. The fix changes the return value to `db_name or self._current_datasource`, preserving the user's database name.

## Test Cases

Manual verification:
1. Run `datus-agent run --datasource X --task "..." --task_db_name Y` where Y is a valid database name that differs from the datasource name
2. Verify the workflow logs show `database_name='Y'` instead of `database_name='X'`

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved datasource identification when an explicit database name is provided, ensuring the selected name is preserved in fallback scenarios.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database name/type resolution when a provided database name doesn’t match a known datasource key—now preserves the caller’s database name while deriving the database type from the currently selected datasource (including collision and single-datasource fallback scenarios).

* **Tests**
  * Updated and added unit tests covering: datasource key resolution, custom database name handling, deriving type from the current datasource (not the name), and fallback when no current datasource is set.

### DIFF
--- a/datus/configuration/agent_config.py
+++ b/datus/configuration/agent_config.py
@@ -1794,7 +1794,7 @@ class AgentConfig:
         if db_name and db_name in datasources:
             return db_name, datasources[db_name].type
         if self._current_datasource and self._current_datasource in datasources:
-            return self._current_datasource, datasources[self._current_datasource].type
+            return db_name or self._current_datasource, datasources[self._current_datasource].type
         if len(datasources) == 1:
             cfg = list(datasources.values())[0]
             return next(iter(datasources)), cfg.type
--- a/datus/configuration/agent_config.py
+++ b/datus/configuration/agent_config.py
@@ -1791,10 +1791,14 @@ class AgentConfig:
 
     def current_db_name_type(self, db_name: str) -> tuple[str, str]:
         datasources = self.services.datasources
-        if db_name and db_name in datasources:
-            return db_name, datasources[db_name].type
+        if db_name:
+            if self._current_datasource and self._current_datasource in datasources:
+                return db_name, datasources[self._current_datasource].type
+            if len(datasources) == 1:
+                cfg = list(datasources.values())[0]
+                return db_name, cfg.type
         if self._current_datasource and self._current_datasource in datasources:
-            return db_name or self._current_datasource, datasources[self._current_datasource].type
+            return self._current_datasource, datasources[self._current_datasource].type
         if len(datasources) == 1:
             cfg = list(datasources.values())[0]
             return next(iter(datasources)), cfg.type
--- a/tests/unit_tests/configuration/test_configuration_load.py
+++ b/tests/unit_tests/configuration/test_configuration_load.py
@@ -188,3 +188,17 @@ def test_get_db_name_type_with_custom_db_name(agent_config: AgentConfig):
     db_name, db_type = agent_config.current_db_name_type(db_name="my_custom_db")
     assert db_name == "my_custom_db"
     assert db_type == DBType.SQLITE
+
+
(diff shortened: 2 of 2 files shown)


pika-Ysw:
邮件已收到~

datus/configuration/agent_config.py
```diff
@@ -1794,7 +1794,7 @@ def current_db_name_type(self, db_name: str) -> tuple[str, str]:
         if db_name and db_name in datasources:
             return db_name, datasources[db_name].type
         if self._current_datasource and self._current_datasource in datasources:
-            return self._current_datasource, datasources[self._current_datasource].type
+            return db_name or self._current_datasource, datasources[self._current_datasource].type
```
stukid: Could we remove the `db_name in datasources` interpretation and always derive the database type from the selected datasource? `db_name` comes from `--task_db_name`, so it represents a physical database name rather than a datasource key. If it happens to match another configured datasource name (for example, current datasource `bird_sqlite` with task database `starrocks`), the earlier branch returns `database_type="starrocks"` while `SqlTask.datasource` remains `bird_sqlite`, leaving the dialect and connector inconsistent. Please preserve the explicit database name, derive the type from `self._current_datasource`, and add a regression test for this name-collision case.

pika-Ysw: Thanks for the review! Updated as suggested. 1. Removed the db_name in datasources check - db_name is now always treated as a physical database name, not a datasource key. 2. The type is always derived from self._current_datasource. 3. Added a regression test for the name-collision case: when current_datasource=bird_school (type SQLITE) and db_name=starrocks (also a configured datasource), it now correctly returns (starrocks, DBType.SQLITE).

stukid:
@pika-Ysw thanks for contributing! I'll merge it.

PR #1138, opened 2026-07-13
[Enhancement] Declare AgentConfig.prompt_manager as a typed runtime attribute

Louis-Law:
## Why

`get_prompt_manager()` documents `agent_config.prompt_manager` as **entry #1** in its resolution order:

```
Resolution order:
1. ``agent_config.prompt_manager`` if already attached
2. A new ``PromptManager`` bound to ``agent_config`` (and its path_manager)
3. Default ``PromptManager()``
```

…but `AgentConfig` never defined the attribute. The extension point added in #518 (*"Replace PromptManager singleton with factory pattern for SaaS multi-tenant isolation"*) only existed in a `getattr()` lookup — hosts had to graft the attribute on from outside, and nothing told a reader, an IDE, or a type checker that it was there at all.

The SaaS backend is the motivating consumer. `PromptManager` derives its template dir from `AgentConfig.home` (`DatusPathManager.template_dir == home/"template"`), but the backend's `home` is the *per-project* dir while prompt templates are a **workspace-level** asset shared across the projects in that workspace. It therefore attaches a `PromptManager` anchored one level up. That works today purely because Python lets you set an undeclared attribute — which is not a contract anyone can rely on.

## Solution

Declare it in `AgentConfig.__init__`:

```python
self.prompt_manager: Optional["PromptManager"] = None
```

`None` means *"derive the template dir from `home`"* — every CLI run. A host whose templates do not live under `home` assigns its own manager. No behavior changes: `get_prompt_manager()` already treats `None` as "fall through to `path_manager`", which is exactly what an unset attribute did before.

`path_manager` is annotated the same way in `_set_path_manager`, so the two runtime attributes now read alike.

### Why an instance attribute and not a dataclass field

This is the key design decision, and it is deliberate.

`AgentConfig` is a `@dataclass`, so a *declared field* lands in `dataclasses.asdict()` — which is precisely what `DatusService.compute_fingerprint()` hashes:

```python
payload = dataclasses.asdict(agent_config)
serialized = json.dumps(payload, sort_keys=True, default=str)
return hashlib.sha256(serialized.encode("utf-8")).hexdigest()
```

A `PromptManager` has no custom `__repr__`, so `default=str` would stringify it to `<PromptManager object at 0x...>` — a **new memory address on every rebuild**. Promoting `prompt_manager` to a field would therefore churn the fingerprint on every config rebuild and evict the cached `DatusService` mid-session. Keeping it an instance attribute (exactly like `path_manager`) keeps it out of `asdict()` entirely.

The same constraint already governs the `model_extras` normalization a few lines below, which explicitly notes it normalizes to a plain `dict` "so `dataclasses.asdict` serializes cleanly into the AgentConfig fingerprint".

## Test Cases

New `TestPromptManagerAttribute` in `tests/unit_tests/configuration/test_agent_config.py` (5 unit tests, no external deps):

- `test_defaults_to_none` — unset means "derive from `home`".
- `test_get_prompt_manager_falls_back_to_path_manager_when_unset` — the CLI path is unchanged.
- `test_attached_manager_overrides_the_home_derived_one` — an attached manager wins, and its template dir is *not* the `home`-derived one.
- `test_excluded_from_asdict_so_the_fingerprint_stays_stable` — pins the design decision above: `prompt_manager` stays out of `dataclasses.asdict()` and out of `__dataclass_fields__`, and attaching one leaves the payload byte-identical. A future refactor that promotes it to a field fails here instead of silently degrading the service cache.
- `test_survives_deepcopy` — hosts clone the config per request (`chat_task_manager` does `copy.deepcopy(agent_config)` before every agentic loop); the clone keeps its own manager.

Gates run locally on this branch:

- `ruff format` / `ruff check` — clean.
- `ci/audit_tests.py --diff-only origin/main` — **P0=0, P1=0**.
- `ci/run-pr-tests.py origin/main` — impacted unit tests **412 passed** (exit 0); diff coverage 66%.
- `tests/unit_tests/configuration/` + `tests/unit_tests/prompts/` — **489 passed**.

One pre-existing failure in the acceptance suite, unrelated to this change: `tests/integration/storage/test_doc_search.py::TestSearchTool::test_delete_docs_by_version` panics inside LanceDB with `DecodeError: expected type URL "/lance.encodings.PageLayout" (got "/lance.encodings21.PageLayout")` — a lance encoding-version mismatch against the pre-built test data. Verified it fails identically on a clean `origin/main` (96f45ca7) with none of this branch's changes applied.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved agent configuration handling for prompt managers.
  * Prompt managers now default safely, support explicit overrides, and remain independent when configurations are copied.

* **Tests**
  * Added coverage for default behavior, custom template paths, configuration serialization, and deep-copy behavior.

### DIFF
--- a/datus/configuration/agent_config.py
+++ b/datus/configuration/agent_config.py
@@ -7,7 +7,7 @@ import os
 import re
 from dataclasses import asdict, dataclass, field, fields
 from pathlib import Path
-from typing import Any, Dict, List, Mapping, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Mapping, Optional, Union
 
 from datus.configuration.node_type import NodeType
 from datus.observability.config import ObservabilityConfig
@@ -19,6 +19,10 @@ from datus.utils.constants import DBType
 from datus.utils.exceptions import DatusException, ErrorCode
 from datus.utils.loggings import get_logger
 from datus.utils.path_utils import get_files_from_glob_pattern
+
+if TYPE_CHECKING:
+    from datus.prompts.prompt_manager import PromptManager
+    from datus.utils.path_manager import DatusPathManager
 
 # Regex for validating platform/identifier names (no special chars that break paths)
 _SAFE_NAME_RE = re.compile(r"^[a-zA-Z0-9_\-]+$")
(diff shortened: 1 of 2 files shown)


Louis-Law:
Superseded by #1139 — same change, but the branch belongs on my fork rather than pushed straight to the org repo. Closing this one.

PR #1158, opened 2026-07-17
[BugFix] Deliver @-context to all subagents + replay it in chat history

arnozhang:
## Why

@Table / @Metric / @Sql references (`table_paths` / `metric_paths` / `sql_paths`) sent with a chat turn never reached the **semantic subagents** (`gen_metrics`, `gen_semantic_model`, `gen_table`, `gen_job`):

- `SemanticNodeInput` had no fields to carry them, so `node_factory.create_node_input` silently dropped the resolved context, and the `task()`-spawned subagent path (`sub_agent_task_tool._build_node_input`) forwarded none of it.
- Result: `/gen_metrics` with `table_paths` set still re-asked the user "which table?" — the reported bug.

The same gap affected `ask_metrics` / `gen_report` / `gen_visual_report` / `gen_visual_dashboard` / `gen_dashboard` / `scheduler` / `sql_summary` / `explore` inputs. On the consumption side, the `DeliverableAgenticNode` and `BaseVisualArtifactAgenticNode` `_build_enhanced_message` overrides bypassed the base at-context rendering entirely.

Separately, chat **history could not re-render** the references a user attached — they were only ever folded into the enhanced-prompt text and stripped on read.

## Solution

- **One carrier, no leaks.** New `AtContextInput` base (`datus/schemas/at_context.py`) declares `schemas` / `metrics` / `reference_sql` / `external_knowledge`; every subagent input inherits it. `ChatNodeInput` / `GenSQLNodeInput` drop their duplicated fields.
- **One choke point.** A single `apply_at_context` helper is called by both node-input funnels (`create_node_input` and `_build_node_input`); the task tool now inherits the parent turn's @-context (mirroring `_parent_db_context`).
- **One renderer.** `AgenticNode._render_at_context_parts` is the sole place that formats @-context into the prompt, shared by the base `_build_enhanced_message` and the deliverable/visual overrides — a new reference kind renders everywhere by editing one method.
- **History replay.** Each turn's raw reference paths are persisted in a session side table (`user_message_context`, keyed by `user_turn_number`, never fed back to the LLM) and echoed by `GET /api/v1/chat/history` on `SSEMessagePayload.at_context`.

Tradeoff: `knowledge_paths` are persisted/echoed for display, but not resolved into `external_knowledge` (no knowledge completer exists yet) — noted for follow-up.

## Test Cases

- `tests/unit_tests/schemas/test_at_context.py` — anti-leak guard (every subagent input inherits `AtContextInput` and carries all 4 fields), `apply_at_context` set/skip/preserve semantics, `historical_sql` alias round-trip.
- `tests/unit_tests/models/test_session_manager.py::TestUserMessageContext` — save/get round-trip, correct per-turn mapping (no cross-turn bleed), empty-context no-op, and missing-side-table graceful read.
- Full unit suite green (16490 passed; 2 pre-existing failures require local `bird_sqlite` benchmark data and are unrelated to this change).

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
- [ ] release/0.3.8
- [ ] release/0.3.9
## Summary by CodeRabbit

* **New Features**
  * Added support for preserving and replaying referenced tables, metrics, SQL, and external knowledge across chat history.
  * Ensured referenced context is consistently available to agent and subagent workflows.
  * Added support for legacy historical SQL references.

* **Bug Fixes**
  * Improved context mapping to the correct user message and graceful handling when context is unavailable.

* **Tests**
  * Added coverage for context propagation, persistence, replay, defaults, aliases, and unsupported fields.

## Summary by CodeRabbit

* **New Features**
  * Added support for preserving and replaying `@`-referenced tables, metrics, SQL, and business knowledge in chat history.
  * Improved context resolution with subject-aware metric and SQL details.
  * Added graceful fallback hints when referenced information cannot be loaded.
  * Enhanced generated prompts with clearer, structured context sections and lookup guidance.

* **Bug Fixes**
  * Prevented referenced context from being lost between turns or incorrectly attached to other messages.
  * Improved handling of ambiguous, incomplete, and legacy reference paths.

### DIFF
--- a/datus/agent/node/agentic_node.py
+++ b/datus/agent/node/agentic_node.py
@@ -33,7 +33,6 @@ from datus.prompts.prompt_manager import get_prompt_manager
 from datus.schemas.action_history import ActionHistory, ActionHistoryManager, ActionRole, ActionStatus
 from datus.schemas.base import BaseInput, BaseResult
 from datus.utils.exceptions import DatusException, ErrorCode
-from datus.utils.json_utils import to_str
 from datus.utils.loggings import get_logger
 from datus.utils.message_utils import build_structured_content
 from datus.utils.node_utils import build_database_context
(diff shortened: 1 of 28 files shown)


arnozhang:
### Code-review follow-up (pushed)

- **Turn-attribution hardening.** `save_user_message_context` now gates the write on a `previous_turn_number` snapshot taken before the run — it only persists when `MAX(user_turn_number)` strictly increased. This removes the reliance on the implicit "one user turn per run" invariant: a run that adds no user turn (or a failed/cancelled one) can no longer mis-attach @-context onto the previous turn's bubble. Added `test_guard_skips_when_no_new_turn`.
- **branch_id.** Documented that `user_turn_number` is session-global monotonic (assigned by the SDK across all branches), so a bare `MAX` is correct even after rewind/fork — no `branch_id` filter needed.

### Behavior note (disclosure)

The prompt-assembly refactor moves the `external_knowledge` ("MUST use these business logic") block from *before* the datasource/DB-context line to *after* it (now grouped with the other @-context parts in `_render_at_context_parts`). All parts are joined into one `<system_reminder>` block, so this is a cosmetic ordering change with no semantic effect — flagging it since it wasn't in the original description.

### Known tradeoff (unchanged)

`knowledge_paths` are persisted and echoed for chip display but not yet resolved into `external_knowledge` (no knowledge completer exists). Tracked as follow-up; noting the potential UX expectation that a @Knowledge chip implies the knowledge is in effect.

PR #1185, opened 2026-07-22
[Feature] Author subject-tree metrics through the semantic adapter

arnozhang:
## Why

`/api/v1/subject/metric` (and create/edit/delete) reconstructs metric YAML from the LanceDB projection. As a result **OSI-format metrics are returned in MetricFlow shape** (`type: aggregate` + `locked_metadata`), losing OSI-native fields (`expression.dialects`, `ai_context`, `custom_extensions`), and edit/create/delete write MetricFlow-shaped files that corrupt or miss OSI semantic-model documents.

We treat the `subject/semantic_models` YAML files as the **source of truth**, with LanceDB demoted to a derived index. This PR wires the explorer endpoints to the semantic adapter's authoring API (added in datus-semantic-adapter) so reads/writes go through the format-correct adapter (OSI / MetricFlow) instead of format logic hard-coded in the backend.

Depends on the datus-semantic-adapter authoring API (`read/write/delete/validate_metric_source` on `BaseSemanticAdapter`, implemented for OSI and MetricFlow). That surface is backend-only — not added to `SemanticTools.all_tools_name()`, never exposed as an agent/LLM tool.

## Solution

New `ExplorerService` helpers: `_semantic_adapter()`, `_is_osi_authoring()`, `_metric_name_from_yaml()`, `_sync_osi_file_to_kb()`, `_osi_metric_exists()`, `_author_osi_metric()`, `_rollback_osi_write()`.

- **get_metric**: the KB row stays the **access-control gate** — it still enforces the full `subject_path` match and sub-agent scoping, so wrong-path or cross-agent reads keep 404-ing. Only the returned *content* comes from the file (via `adapter.read_metric_source`), which is native-format and accurate; falls back to the KB reconstruction when authoring is unsupported. Note this means MetricFlow reads now also come from the file (still `{metric: ...}`-shaped, identical to the old `_metric_db_to_yaml` output) — intentional, since the file is more accurate than the lossy KB rebuild.
- **create/edit_metric**: when the adapter is OSI, author in place inside the `semantic_model.metrics` list, preserving datasets and sibling metrics and recording `subject_path` in DATUS `custom_extensions`, then re-index the changed file. To keep the file and KB consistent, a failed KB re-sync **rolls back**: a create is deleted, an edit is restored to its pre-edit snapshot. MetricFlow keeps its existing path unchanged.
- **delete (metric)**: for OSI, remove from the semantic-model file first; a **real write failure fails the request**, while an already-absent metric (file/KB drift) is tolerated so we still drop the stale KB row. This prevents a swallowed file-delete error from dropping the KB row while the file keeps the metric (which would "revive" on the next re-index).
- KB-sync failures map to `TOOL_EXECUTION_ERROR`. Adapter file I/O + KB embedding run via `asyncio.to_thread` to avoid blocking the event loop.

Multi-sub-agent isolation: access scoping is unchanged because the KB gate (`get_metrics_detail`, which applies `_sub_agent_conditions()`) still fronts every read; the adapter is only reached after that check passes.

## Test Cases

Added `TestExplorerServiceOSIAuthoring` in `tests/unit_tests/api/services/test_explorer_service.py`, driving `ExplorerService` end-to-end against a real `DatusOSIAdapter` over temp files (11 cases):
- get_metric returns OSI-native YAML; falls back to KB reconstruction when authoring unsupported; 404s when the KB row is missing (scope gate)
- create writes into the semantic-model file and re-indexes the KB; rejects invalid YAML without writing; rolls back the created metric when KB re-sync fails
- edit updates in place preserving datasets; restores the previous content when KB re-sync fails
- delete removes the metric from the OSI file; fails the request (and keeps the KB row) on a real source-file write failure
- create fails cleanly when the adapter is unavailable

`datus_semantic_osi` is guarded with `pytest.importorskip`. Full file: 82 passed, no regressions to existing MetricFlow tests.

## Summary by CodeRabbit

* **New Features**
  * Metrics can now be authored and retrieved using semantic-adapter-managed “source of truth,” including OSI semantic models.
  * Metric create/edit/delete operations are now adapter-aware with knowledge-base re-sync.
* **Bug Fixes**
  * Prevented metrics from being incorrectly marked stale when upserting/resyncing with matching IDs.
  * Improved fallback and not-found behavior when adapter/OSI sources can’t be read.
  * Ensured validation and re-sync failures roll back OSI/adapter changes.
* **Tests**
  * Added dedicated OSI metric authoring and adapter-based metric authoring coverage; updated storage tests for identity/conflict cases.
* **Chores**
  * Updated dependency constraints for CI and semantic core.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
- [ ] release/0.3.8
- [ ] release/0.3.9

### DIFF
--- a/datus/api/services/explorer_service.py
+++ b/datus/api/services/explorer_service.py
@@ -135,14 +135,16 @@ class ExplorerService:
         )
 
     @staticmethod
-    def _metric_exists_in_source(adapter, metric_name: str, parent_path: List[str]) -> bool:
-        """Whether the metric still reads from the source file (used after a
-        delete error to tell a real write failure from an already-absent metric)."""
-        try:
-            adapter.read_metric_source(metric_name, subject_path=parent_path)
-            return True
-        except Exception:  # noqa: BLE001 - any read failure means "not present"
-            return False
+    def _is_metric_absent_error(exc: Exception) -> bool:
+        """Whether a delete failure means the metric simply isn't in the source
+        file (benign file/KB drift) rather than a real I/O / lock / parse failure.
+
+        Both adapters raise a not-found error whose message contains
+        ``was not found`` (``FileNotFoundError`` for MetricFlow, the OSI error
+        class for OSI). Anything else is a real failure and must not be treated
+        as "already gone", or we would drop the KB row while the file still
+        holds the metric (the file is the source of truth)."""
+        return isinstance(exc, FileNotFoundError) or "was not found" in str(exc)
 
     def _author_metric(
         self,
@@ -203,9 +205,7 @@ class ExplorerService:
         try:
             # Empty parent_path (root-level metric) means "no categorization" —
             # pass None so the adapter does not inject an empty subject tag.
-            mutation = adapter.write_metric_source(
-                name, metric_yaml, subject_path=(parent_path or None), create=create
-            )
+            mutation = adapter.write_metric_source(name, metric_yaml, subject_path=(parent_path or None), create=create)
         except Exception as e:  # noqa: BLE001 - surface adapter write/validation errors
             logger.error(f"Failed to write metric source: {e}")
             return Result[dict](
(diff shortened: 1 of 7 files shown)


arnozhang:
Follow-up fix pushed (`5aba0729`): editing an OSI metric re-syncs the whole semantic-model file into the KB, and when the KB already held that metric under a different `semantic_model_name` (model renamed / prior sync under a different model), the upsert conflict guard raised *"Metric name conflict within datasource ... existing metric id '...' has a different 'semantic_model_name'"* and the edit was rolled back.

Root cause: metric ids are name-derived (`metric:<name>`), so a same-name write is the *same* identity — a legitimate overwrite on re-sync (the YAML file is the source of truth), not a collision. `_find_stale_existing_metric_ids` now skips the conflict/stale check when the existing row's id equals the incoming id (the batch upsert overwrites in place), and still raises only for a genuinely distinct id sharing the name. Added regression tests for both paths.

PR #1256, opened 2026-08-06
[Refactor] ask_metrics: stop inferring window-metric dependencies in the node

stukid:
## Why

`query_metrics` expands a requested window metric into a bundle — its base metric plus, for averages, the matching window row-count metric — so the answer carries the underlying series and the count the average was computed over.

**The need is real.** A 7-day moving average without the raw series can't tell you whether a dip is real or smoothed away, and without the window count you can't see that the first few points were averaged over 3 rows instead of 7. Deriving it *in the node* is the problem.

Nothing declares those relationships, so the node reverse-engineers them:

- an explicit name under any of **four** accepted keys (`base_metric`, `base_metric_name`, `source_metric`, `source_metric_name`) — four spellings means there was never a convention;
- failing that, a set intersection over **measure expression strings**, normalised by lowercasing and stripping whitespace — `COUNT(*)` and `count(1)` miss;
- window counts matched by requiring `dataset`, `time_dimension`, `window` and `grain_to_date` to be pairwise equal **as free text**.

Each pass scans the whole catalog, which is why `_metric_catalog()` pages up to 5000 metrics into a node-lifetime cache.

**Every field it reads is MetricFlow-shaped.** The Dosi adapter exposes

```python
MetricDefinition(..., metadata={"datasets": [...]})
```

and nothing else — no `window`, `window_aggregation`, `grain_to_date`, `time_dimension` or `base_metric`. So `_is_window_metric` is always `False` and `_window_metric_bundle` returns its input unchanged. Verified directly:

```
metadata          : {'datasets': ['orders']}
_is_window_metric : False
bundle            : ['order_count_7d_avg']
```

The inference and the 5000-metric fetch are dead weight on that path — the same shape as the query preflight removed in #1251.

Where a derived metric's dependencies belong is the semantic model. OSI `Metric` is `#[serde(deny_unknown_fields)]` but reserves `custom_extensions` and `ai_context` extras, so a model can carry them without the agent guessing.

## Solution

Remove the inference (`_expand_metric_dependencies`, `_expand_window_metrics`, `_window_metric_bundle`, `_find_window_base_metric`, `_matching_window_count_metrics`, `_window_metric_needs_window_count`, `_is_window_metric`, `_is_base_metric_candidate`, `_metric_measure_signature`, `_normalize_metric_expression`, `_metric_metadata`, `_metadata_text`, `_first_metadata_text`, `_append_unique`) together with `_metric_catalog()` and its cache. Requested metrics are normalised and forwarded as-is.

The tool docstring now tells the model to request the underlying series or the window row count explicitly when the answer needs them, instead of promising an expansion that no longer happens.

## Test Cases

- `test_query_metrics_forwards_requested_metrics_unchanged` / `test_query_metrics_forwards_multiple_metrics_in_order` replace the two bundle-expansion tests: requested metrics reach the adapter verbatim
- `tests/unit_tests/agent/node/` + `tests/unit_tests/tools/func_tool/`: **3772 passed, 1 skipped**

Net **-298 lines**.

## Note

Stacked on the same area as #1255 (removing the rejection cache) but touches disjoint code — either can land first.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
- [ ] release/0.3.8
- [ ] release/0.3.9

## Summary by CodeRabbit

* **Behavior Changes**
  * Metric queries now request only the metrics explicitly provided.
  * Derived and underlying metrics must be requested separately when needed.
  * Requested metrics are forwarded in the same order supplied.
* **Bug Fixes**
  * Prevented automatic expansion of metric dependencies, reducing unexpected or unnecessary metric results.
  * Improved consistency for rolling-average and running-extrema metric queries.

### DIFF
--- a/datus/agent/node/ask_metrics_agentic_node.py
+++ b/datus/agent/node/ask_metrics_agentic_node.py
@@ -87,7 +87,6 @@ class AskMetricsAgenticNode(AgenticNode):
         self.subject_tree_metric_entries: List[Dict[str, Any]] = []
         self.subject_tree_mode: str = "none"
         self.subject_tree_prompt: str = ""
-        self._metric_catalog_cache: Optional[Dict[str, Dict[str, Any]]] = None
         self._failed_query_signatures: set[tuple] = set()
         self._selected_final_metric_result_id: Optional[str] = None
         self.startup_error: Optional[str] = None
@@ -440,10 +439,10 @@ class AskMetricsAgenticNode(AgenticNode):
         N/Bottom N, also pass order_by so the truncation has stable business
         meaning.
 
-        For fixed period-over-period metrics such as month-over-month or
-        year-over-year, query the dedicated catalog metric directly. For window
-        metrics, AskMetrics expands the request to include related executable
-        metrics when they are already present in the catalog.
+        For derived metrics such as month-over-month, year-over-year or moving
+        averages, query the dedicated catalog metric directly. When the answer
+        also needs the underlying series or the window's row count, request
+        those metrics explicitly alongside it.
 
         Joined dimensions always use matched-rows join semantics: fact rows
         that cannot match the requested dimension are excluded. Unmatched-fact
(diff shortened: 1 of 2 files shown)


stukid:

PR #1257, opened 2026-08-06
[BugFix] ask_metrics: align join-semantics guidance with the model declaration

stukid:
## Why

The `query_metrics` docstring and the ask_metrics system prompt assert a fixed join behaviour:

> Grouping by a joined dimension **always** uses matched-rows join semantics: fact rows that cannot match the requested dimension are excluded. **No other join semantics are available.**

and forbid the only way to audit those rows:

> **Never** emulate it with a `where` filter such as `<dimension> IS NULL`: matched-rows semantics removes those rows, so the query would return a misleadingly empty result.

That was written in #1203 against the OSI adapter's `match_only` default, with an explicit tradeoff attached:

> …degrade to explicit refusal or a labeled matched-rows answer **until modeling-time declarations (D-JOIN/D-FILL) are consumed by the adapter**

**They are consumed now.** dosi-engine implements D-JOIN (`join_type` on a relationship) and D-FILL (`fill_nulls_with` on a metric) as documented `DATUS` custom extensions, and `dosi-compiler/src/ext.rs` states the default:

```rust
/// Default (no `datus` extension, or no `join_type`) is `Left`.
```

with an oracle test (`inner_join_type_drops_orphans`) holding the `inner` case, and a user-facing guide:

> By default dosi-engine keeps them (a `LEFT JOIN`): the unmatched revenue lands in a row where `status` is blank (NULL). That is the right answer for **reconciliation**.

So on the Dosi path unmatched facts are **kept** — the opposite of what the prompt claims. Two consequences:

1. the model refuses unmatched-fact audits that the engine supports out of the box;
2. the `<dimension> IS NULL` filter it is explicitly forbidden from using is exactly the correct one there.

And since the two adapters default differently (OSI `match_only`, Dosi `Left`), **no hard-coded assertion can be true for both**.

## Solution

Stop asserting the behaviour; describe how to read it off the result.

| | |
|---|---|
| Unmatched facts | A row with an empty/NULL joined dimension holds them — report as unmatched, don't drop it or treat the blank as a data error |
| Audit request | Filter on `<dimension> IS NULL`. If that returns nothing while `join_policy_filtered_rows` > 0, the model excludes unmatched facts before they can be filtered — say so instead of reporting zero |
| Disclosure rule | **Kept** — it consumes metadata the adapter already emits rather than declaring a capability |
| Absent dimension values | Still never returned; and whether a blank metric value is filled stays a model decision, so the answer must not substitute zero on its own |

## Test Cases

- `test_query_metrics_preserves_join_filtered_rows_metadata` still guards the metadata contract the disclosure rule depends on — unchanged
- `tests/unit_tests/agent/node/` + `tools/func_tool/` + `prompts/`: **236 passed**
- Full `tests/unit_tests/`: **9356 passed, 1 skipped, 1 xfailed**. The single failure (`plugins/test_runtime_context.py::test_bash_pipeline_scopes_context_to_plugin_segment`) reproduces on unmodified `upstream/main` — verified by stashing this change — and is unrelated.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
- [ ] release/0.3.8
- [ ] release/0.3.9

## Summary by CodeRabbit

* **Documentation**
  * Clarified how metric results handle unmatched facts when dimensions are joined.
  * Results may group unmatched facts under a blank or NULL dimension value, or exclude them according to relationship policies.
  * Excluded rows are disclosed in result metadata, including the exclusion count.
  * Dimension values with no matching facts are not shown.
  * Blank metric values remain blank rather than being replaced with zero.


stukid:

PR #1258, opened 2026-08-06
[Refactor] ask_metrics: use the semantic tool's query_metrics directly

stukid:
> **Stacked on #1255, #1256, #1257** — those three are merged into this branch, so the diff shown here includes them. Review/merge them first; this PR's own change is the last commit.

## Why

The node replaces `SemanticTools.query_metrics` with its own bound method at tool-registration time. Once the rejection cache (#1255) and the window-metric inference (#1256) are gone, what remains is a forwarder: normalise → check empty → call the tool → apply column aliases.

Nothing left in it belongs to the node.

**1. Normalisation and the empty check duplicate the tool** — down to a byte-identical error message. And the node's `_normalize_string_list` is the *weaker* of the two:

| | tool `_normalize_name_list` | node `_normalize_string_list` |
|---|---|---|
| literal `"null"` / `"None"` | stripped via `normalize_null` | kept as a metric name |
| `set` argument | accepted | returns `[]` |

**2. The docstring paragraphs are already in the system prompt.** All three — result handling, derived metrics, join semantics — have a matching rule in `ask_metrics_system_1.0.j2` (lines 47, 51, 41). They describe tool-layer and adapter-layer behaviour, not anything specific to this node.

**3. Overriding the docstring lost information.** The node's `Args` is a shortened copy, so the AskMetrics model alone could not see:

- the legal `time_granularity` values (`'day', 'week', 'month', 'quarter', 'year'`)
- the relative time syntax (`-7d`)
- that `dimensions` come from `get_dimensions` and `path` from `list_subject_tree`
- **`order_by`: "Do NOT use 'asc'/'desc' keywords"** — a warning the tool spells out explicitly

**4. The column aliases are not stable.** `customers__region` shortens to `region` only when that suffix is unique in the result *and* doesn't collide with an existing column — so the same dimension is named differently depending on what else was grouped:

```
['customers__region', 'revenue']                    ->  ['region', 'revenue']
['customers__region', 'stores__region', 'revenue']  ->  unchanged
['customers__region', 'region', 'revenue']          ->  unchanged
```

`update_context` applies these aliases to the **delivered CSV**, so the user's column headers shift with query shape too. `metric_time__*` is excluded from aliasing entirely, leaving two naming styles in one result. A display name should come from the semantic model, stably — not be reverse-engineered per result set.

## Solution

Delete the wrapper and the alias helpers, and stop substituting the tool in `_setup_specific_tool_method` / `_add_available_tools`. AskMetrics mounts `SemanticTools.query_metrics` like every other node.

Side effect worth noting: the tool's `metrics` parameter has no default, so the generated schema now marks it **required** — the node's signature defaulted it to `None` and rejected the empty case at runtime instead.

`select_final_metric_result`, subject-tree handling and `update_context` stay — those are genuinely node-level.

## Test Cases

Removed 14 node tests that exercised the wrapper's forwarding. Their coverage is held by `tests/unit_tests/tools/func_tool/test_semantic_tools.py`, which has **21** `query_metrics` tests including empty-metrics rejection, argument normalisation and full parameter pass-through.

`tests/unit_tests/agent/` + `tools/func_tool/` + `api/`: **5209 passed, 1 skipped**.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
- [ ] release/0.3.8
- [ ] release/0.3.9

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved joined-dimension results to accurately handle unmatched rows and relationship-specific filtering.
  * Preserved blank metric values for groups without matching facts.
  * Query results now retain their original column names and values.

* **Documentation**
  * Clarified that results are complete by default.
  * Documented using `limit` for explicit row limits and `order_by` for Top/Bottom N queries.


stukid:

PR #1301, opened 2026-08-18
[BugFix] List a schema-only dialect's catalog instead of returning nothing

arnozhang:
## Symptom

Every Oracle datasource came back with an empty catalog:

```
GET /api/v1/catalog/list → 200 {"success":true,"data":{"databases":[]}}
```

The SaaS catalog panel therefore rendered "No tables yet" — while, in the same session and against the same connection, `list_tables` in chat answered correctly:

```
Call Tool list_tables → 2 tables
  DATUS_E2E.E2E_CUSTOMERS, DATUS_E2E.E2E_ORDERS
```

So the connection, the credentials and the metadata queries were all fine; only the listing endpoint produced nothing.

## Cause

`_get_connection_info` resolves what to iterate as **databases**:

```python
if request.database_name:                   db_names = [request.database_name]
elif connector.database_name:               db_names = [connector.database_name]
elif hasattr(connector, "get_databases"):   db_names = connector.get_databases(...)
else:                                       db_names = []
```

Oracle has no database level below the service, and its adapter says so on purpose:

- `OracleConfig` has no `database` field (it is only accepted as an alias for `service_name`), so `connector.database_name` is `''`.
- `OracleConnector.get_databases()` returns `[]` — *"Oracle has no database level below the service; namespace is schema-only."*

Both branches are empty, `db_names == []`, and the `for db_name in db_names:` loop never runs — the schema resolution and table enumeration nested inside it never get a chance. Reproduced directly against a live instance:

```
connector.database_name          : ''
connector.get_databases()        : []
connector.get_schemas()          : ['BAASSYS', 'DATUS_E2E', 'DBSFWUSER', ...]
connector.get_tables(DATUS_E2E)  : ['E2E_CUSTOMERS', 'E2E_ORDERS']
db_names it will iterate         : []  -> DatabaseInfo entries: 0
```

This is not Oracle-specific: any dialect whose only namespace is the schema falls through the same hole.

## Fix

When nothing else produced a root **and** the dialect declares no `database` namespace, report its **schemas as the roots**: `name=<schema>`, `schema_name=None`.

A table then addresses as `schema.table` — exactly the identifier such a dialect accepts, and the shape its own `parse_*_identifier` produces. That matters because the alternative fixes are worse: making the adapter return the service/PDB as a "database" would push `SERVICE.SCHEMA.TABLE` into consumers that build identifiers from this tree, which Oracle cannot resolve, and it is precisely why the adapter keeps `database_name` empty.

`schema_name` is left `None` because the entry is already the root; setting it too would render the same name twice in a catalog tree.

Both guards matter:
- **no roots yet** — a dialect that does report a database is untouched.
- **no `database` capability** — so this cannot flatten a PostgreSQL-shaped catalog.

Per-schema failures still degrade to that one schema (`connection_status: connected` plus the reason), rather than hiding its siblings.

## Verification

`tests/unit_tests/api/services/test_database_service.py` — **61 passed**. `ruff format --check` / `ruff check` clean.

Four new cases: schemas become the roots (names, `schema_name is None`, inlined sorted tables, `current` on the connector's own schema); one unreadable schema does not hide the others; `request.schema_name` narrows the listing without calling `get_schemas`; and a regression case pinning that a dialect **with** a database level keeps its nested `database > schema` shape.

End to end against a real Oracle 23ai container, through `build_database_entry` → `DBManager` → the real `OracleConnector`:

```
before: entries: 0
after : entries: 1
        name='DATUS_E2E' schema_name=None status=connected tables_count=2
        tables=['E2E_CUSTOMERS', 'E2E_ORDERS']
```

Found while verifying Oracle support end to end in Datus-saas (its data source, project and chat all worked; only this panel was empty).

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
- [ ] release/0.3.8
- [ ] release/0.3.9
- [x] release/0.4.0
## Summary by CodeRabbit

* **New Features**
  * Improved database browsing for systems organized by schemas rather than databases.
  * Schemas are now displayed as top-level entries with their tables listed in sorted order.
  * Schema-specific errors are shown while keeping the connection marked as available.
  * Existing schema filters, connection details, and database/schema hierarchies remain supported.

## Summary by CodeRabbit

* **New Features**
  * Improved database browsing for systems organized by schemas rather than databases.
  * Schemas are now displayed as top-level entries with their tables listed in sorted order.
  * Existing schema filters, connection details, and database/schema hierarchies remain supported.

* **Bug Fixes**
  * Schema-specific listing errors are isolated and displayed without preventing other schemas from loading.
  * Connections remain correctly marked as available when individual schema listings fail.
  * Current-schema status is now accurately reflected for schema-based catalogs.

### DIFF
--- a/datus/api/services/database_service.py
+++ b/datus/api/services/database_service.py
@@ -226,6 +226,56 @@ class DatasourceService:
             return [_listing_failed(connector.database_name, str(e))]
 
         db_infos: List[DatabaseInfo] = []
+
+        # 1b) A dialect with no database level (Oracle: schema is the only
+        # namespace) yields nothing above — its connector reports no database and
+        # its get_databases() is empty by design. Iterating that dropped the whole
+        # datasource out of the catalog while its tables were perfectly listable.
+        # Its schemas ARE the top level, so report them as the roots: a table then
+        # addresses as schema.table, which is exactly the identifier such a dialect
+        # accepts. Only when nothing else produced a root, so a dialect that does
+        # report a database keeps its existing shape.
+        if not db_names and has_schema and not supports_namespace("database", connector=connector, dialect=dialect):
+            try:
+                schemas = (
+                    [request.schema_name]
+                    if request.schema_name
+                    else connector.get_schemas(
+                        catalog_name=request.catalog_name,
+                        include_sys=request.include_sys_schemas,
+                    )
+                )
+            except Exception as e:
+                logger.warning("Failed to get schemas for schema-only dialect=%s: %s", dialect, e)
+                return [_listing_failed(connector.schema_name, str(e))]
+
+            for schema in schemas:
+                try:
+                    tables = connector.get_tables(catalog_name=catalog_name, schema_name=schema)
+                    tables.sort()
+                except Exception as e:
+                    logger.warning("Failed to get tables for schema=%s dialect=%s: %s", schema, dialect, e)
+                    db_infos.append(_listing_failed(schema, str(e)))
+                    continue
+
+                db_infos.append(
+                    DatabaseInfo(
+                        name=schema,
+                        uri=_get_uri(connector),
+                        type=dialect,
+                        current=(schema == connector.schema_name),
+                        catalog_name=catalog_name,
+                        # Already the root: nesting it under itself would render the
+                        # same name twice in a catalog tree.
+                        schema_name=None,
+                        connection_status="connected",
+                        tables_count=len(tables),
+                        last_accessed=now,
+                        tables=tables,
+                    )
+                )
+            return db_infos
+
         for db_name in db_names:
             if has_schema:
                 # 2) Resolve schemas for this db — a single failing db must not
--- a/datus/api/services/database_service.py
+++ b/datus/api/services/database_service.py
@@ -175,17 +175,24 @@ class DatasourceService:
                 last_accessed=now,
             )
 
-        def _listing_failed(db_name: str, error: str, schema: Optional[str] = None) -> DatabaseInfo:
+        def _listing_failed(
+            db_name: str, error: str, schema: Optional[str] = None, *, current: Optional[bool] = None
+        ) -> DatabaseInfo:
             """The connection works but its objects could not be enumerated.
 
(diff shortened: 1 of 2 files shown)


arnozhang:
Thanks — went through all three against the current code.

**Fixed: `current` on schema-root failures** (`bac95e57`). Valid, and introduced here: `_listing_failed` derives `current` from `db_name == connector.database_name`, which is right for a database root but wrong for the schema roots this PR adds — a schema-only dialect reports no `database_name`, so its own default schema came back `current: false` on failure while the success path immediately below reported `current: true` for the same schema. `_listing_failed` now takes an optional `current` override (default unchanged, so database roots behave exactly as before) and both schema-root failure paths pass the schema comparison. Two cases added: a failing default schema still reads as current, and a `get_schemas()` failure reports the connector's own schema as the current root.

**Fixed: type annotations on the new test helpers** (same commit). The module-level helpers in this file (`_service_with_semantic_adapter`, `_write_model`, `_osi_yaml`) are annotated, so the new ones now match, `_tables` included.

**Skipping: rejecting `request.database_name` for schema-only dialects.** That branch predates this PR — `if request.database_name: db_names = [request.database_name]` is the existing contract, and the new schema-root path is deliberately guarded on `not db_names` so no existing shape changes. Nothing in this repo or in Datus-saas sends `database_name` for such a dialect (the SaaS catalog panel calls `/catalog/list` with no params at all), so this fixes no live path, and choosing between *reject with an error* and *silently map to `schema_name`* is an API-contract decision worth its own change rather than a rider on a bug fix — a caller passing the service name today to label the root would break under either option. Happy to do it as a follow-up if you want a specific behavior.

Verified after the change: `tests/unit_tests/api/services/test_database_service.py` — **63 passed**, `ruff format --check` / `ruff check` clean, and the end-to-end run against a live Oracle container still shows the catalog populated (`DATUS_E2E` with `E2E_CUSTOMERS` / `E2E_ORDERS`) in the Datus-saas panel that was empty before.

PR #1324, opened 2026-08-22
[BugFix] Route scoped table matching through the active connector

### LINKED ISSUE #1317: Fix Hive describe table with cross-database table references
When trying to execute hive sql, describe table tool will be invoked first. However, I find even if my sql file writting correct database name and table name like "dwd.dwd_dim_user_a_day" while agent config setting hive database to be other ones like "ods", this tool always describe ods.dwd_dim_user_a_day first and then failed. Fortunately, agent can auto fix it but this step always happens. 

Image

simonyang08:
## Why

Scoped table patterns are parsed against the primary connector, while table coordinates can be parsed with the connector routed for each request. In a multi-connector deployment, the same namespace token can therefore be interpreted differently by the scope and by the coordinate being checked.

Related to #1317. Depends on Datus-ai/datus-db-adapters#111 for the Hive root-cause fix.

## Current patch

This Draft routes table-level scope-pattern parsing through the active connector so the pattern and coordinate can use the same connector capabilities.

It does not change the Hive adapter's declared capabilities and does not fix #1317 by itself. Review also identified remaining work before this generic change could be merge-ready: database/schema-level scope checks still use primary-connector patterns, the routing test needs a production-realistic behavior failure on the base revision, and routed-pattern caching and fail-closed diagnostics need a bounded design.

## Validation recorded for this branch

- `python -m pytest tests/unit_tests/tools/func_tool/test_database.py -q`: 165 passed.
- `python -m pytest tests/integration/tools/test_mcp_server.py -q`: 52 passed.
- `ruff format --check` and `ruff check` for both changed files: passed.
- `python ci/audit_tests.py --repo-root . --diff-only upstream/main`: P0=0, P1=0.
- `tests/integration/tools/test_func_tools_db.py`: 38 passed with 3 missing-local-datasource setup errors, identical on the base and patched snapshots.

These checks establish repository compatibility but do not establish that the current Draft fixes the published Hive path. The full Windows run recorded 1937 passed and 7 skipped; remaining baseline failures/errors involved Windows symlink privileges, WSL/bash path translation, local encoding/fixtures, and concurrent LanceDB temporary writes.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
- [ ] release/0.3.8
- [ ] release/0.3.9
- [ ] release/0.4.0

### DIFF
--- a/datus/tools/func_tool/database.py
+++ b/datus/tools/func_tool/database.py
@@ -579,6 +579,7 @@ class DBFuncTool:
             return None
         dialect = getattr(self._primary_connector, "dialect", "") or ""
         parsed = parse_table_name_parts(token, dialect)
+        parsed = self._normalize_parsed_namespaces(parsed, self._primary_connector, dialect)
         if not parsed.get("table_name"):
             return None
         values = {
@@ -588,6 +589,22 @@ class DBFuncTool:
             "table": parsed.get("table_name", ""),
         }
         return ScopedTablePattern(raw=token, **values)
+
+    @staticmethod
+    def _normalize_parsed_namespaces(
+        parsed: Dict[str, str], connector: BaseSqlConnector, dialect: str
+    ) -> Dict[str, str]:
+        """Align generic parser output with the connector's namespace model."""
+        if (
+            parsed.get("schema_name")
+            and not parsed.get("database_name")
+            and supports_namespace("database", connector=connector, dialect=dialect)
+            and not supports_namespace("schema", connector=connector, dialect=dialect)
+        ):
+            parsed = dict(parsed)
+            parsed["database_name"] = parsed["schema_name"]
+            parsed["schema_name"] = ""
+        return parsed
 
     def _get_semantic_model(
         self, catalog: str = "", database: str = "", schema: str = "", table_name: str = ""
@@ -879,18 +896,7 @@ class DBFuncTool:
         )
         dialect = getattr(routed_connector, "dialect", "") or ""
         parsed = parse_table_name_parts(raw_name, dialect)
-        # Some database-only adapters (notably Hive) advertise capabilities on
-        # the connector while the generic parser lacks matching registry
-        # metadata. It then right-aligns ``database.table`` as ``schema.table``.
-        # Move that namespace back so the connector default cannot override it.
-        if (
-            parsed.get("schema_name")
-            and not parsed.get("database_name")
-            and supports_namespace("database", connector=routed_connector, dialect=dialect)
-            and not supports_namespace("schema", connector=routed_connector, dialect=dialect)
-        ):
-            parsed["database_name"] = parsed["schema_name"]
-            parsed["schema_name"] = ""
+        parsed = self._normalize_parsed_namespaces(parsed, routed_connector, dialect)
         for field, parsed_field in (
             ("catalog", "catalog_name"),
             ("database", "database_name"),
--- a/datus/tools/func_tool/database.py
+++ b/datus/tools/func_tool/database.py
@@ -879,6 +879,18 @@ class DBFuncTool:
         )
         dialect = getattr(routed_connector, "dialect", "") or ""
         parsed = parse_table_name_parts(raw_name, dialect)
+        # Some database-only adapters (notably Hive) advertise capabilities on
+        # the connector while the generic parser lacks matching registry
+        # metadata. It then right-aligns ``database.table`` as ``schema.table``.
+        # Move that namespace back so the connector default cannot override it.
+        if (
+            parsed.get("schema_name")
+            and not parsed.get("database_name")
+            and supports_namespace("database", connector=routed_connector, dialect=dialect)
+            and not supports_namespace("schema", connector=routed_connector, dialect=dialect)
+        ):
+            parsed["database_name"] = parsed["schema_name"]
+            parsed["schema_name"] = ""
         for field, parsed_field in (
             ("catalog", "catalog_name"),
             ("database", "database_name"),
--- a/datus/tools/func_tool/database.py
+++ b/datus/tools/func_tool/database.py
@@ -573,13 +573,16 @@ class DBFuncTool:
         except Exception:
             return None
 
-    def _parse_scope_token(self, token: str) -> Optional[ScopedTablePattern]:
+    def _parse_scope_token(
+        self, token: str, connector: Optional[BaseSqlConnector] = None
+    ) -> Optional[ScopedTablePattern]:
         token = (token or "").strip()
         if not token:
             return None
-        dialect = getattr(self._primary_connector, "dialect", "") or ""
+        routed_connector = connector or self._primary_connector
+        dialect = getattr(routed_connector, "dialect", "") or ""
         parsed = parse_table_name_parts(token, dialect)
-        parsed = self._normalize_parsed_namespaces(parsed, self._primary_connector, dialect)
+        parsed = self._normalize_parsed_namespaces(parsed, routed_connector, dialect)
         if not parsed.get("table_name"):
             return None
         values = {
@@ -907,10 +910,17 @@ class DBFuncTool:
                 setattr(coordinate, field, self._normalize_identifier_part(parsed[parsed_field]))
         return coordinate
 
-    def _table_matches_scope(self, coordinate: TableCoordinate) -> bool:
+    def _table_matches_scope(self, coordinate: TableCoordinate, connector: Optional[BaseSqlConnector] = None) -> bool:
         if not self._scoped_patterns:
             return True
-        return any(pattern.matches(coordinate) for pattern in self._scoped_patterns)
+        patterns = self._scoped_patterns
+        if connector is not None:
+            patterns = []
+            for pattern in self._scoped_patterns:
+                routed_pattern = self._parse_scope_token(pattern.raw, connector)
+                if routed_pattern is not None:
+                    patterns.append(routed_pattern)
+        return any(pattern.matches(coordinate) for pattern in patterns)
 
     def _filter_table_entries(
         self,
@@ -932,7 +942,7 @@ class DBFuncTool:
                 schema=schema,
                 connector=connector,
             )
-            if self._table_matches_scope(coordinate):
+            if self._table_matches_scope(coordinate, connector=connector):
                 filtered.append(entry)
         return filtered
 
@@ -992,7 +1002,7 @@ class DBFuncTool:
         out_of_scope: List[str] = []
         for name in table_names:
             coordinate = self._build_table_coordinate(raw_name=name, connector=routed_connector)
-            if not self._table_matches_scope(coordinate):
+            if not self._table_matches_scope(coordinate, connector=routed_connector):
                 out_of_scope.append(name)
         return out_of_scope
 
@@ -1403,7 +1413,7 @@ class DBFuncTool:
                 connector=connector,
             )
 
-            if not self._table_matches_scope(coordinate):
+            if not self._table_matches_scope(coordinate, connector=connector):
                 error_msg = f"Table '{table_name}' is outside the scoped context."
                 logger.warning(error_msg)
                 return FuncToolResult(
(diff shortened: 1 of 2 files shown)


datus/tools/func_tool/database.py
```diff
@@ -589,6 +593,22 @@ def _parse_scope_token(self, token: str) -> Optional[ScopedTablePattern]:
...
 
+    @staticmethod
+    def _normalize_parsed_namespaces(
+        parsed: Dict[str, str], connector: BaseSqlConnector, dialect: str
+    ) -> Dict[str, str]:
+        """Align generic parser output with the connector's namespace model."""
+        if (
+            parsed.get("schema_name")
+            and not parsed.get("database_name")
+            and supports_namespace("database", connector=connector, dialect=dialect)
```
stukid, line 604: **This guard never fires for Hive, so #1317 is not fixed.**

`datus-hive` registers with no capabilities at all, in every published version (0.1.0 / 0.1.1 / 0.1.2 / 0.1.3):

```python
connector_registry.register("hive", HiveConnector, config_class=HiveConfig, capabilities=set())
```

and `HiveConnector` (-> `SQLAlchemyConnector` -> `BaseSqlConnector`) does not override `get_effective_capabilities()`, so it falls through to the base implementation, which returns `connector_registry.get_capabilities("hive")` == `set()`.

Therefore `supports_namespace("database", connector=hive_conn, dialect="hive")` is `False`, the condition short-circuits, and the parsed namespaces are returned unchanged.

Reproduced against this branch, modelling the published adapter registration:

```
field_order: ['table']
dwd.dwd_dim_user_a_day -> TableCoordinate(catalog='', database='', schema='dwd', table='dwd_dim_user_a_day')
```

And that coordinate still produces the symptom from the issue: `describe_table` calls `connector.get_schema(database_name='', schema_name='dwd', ...)`, `SQLAlchemyConnector.get_schema` resolves `database_name or self.database_name` -> `'ods'`, then `_sqlalchemy_schema()` returns `database_name or schema_name` -> `'ods'`. The parsed `dwd` is dropped and the tool describes `ods.dwd_dim_user_a_day`.

**Two ways forward, both verified locally:**

1. *Preferred, fixes the root cause.* Have `datus-hive` register `capabilities={"database"}`. With that, all three shapes resolve correctly, and this helper becomes a harmless no-op for Hive because the parser field order is already `[database_name, table_name]`:
   ```
   field_order: ['database', 'table']
   dwd.tbl                -> database='dwd'
   explicit database="dwd" -> database='dwd'
   bare table name         -> database='ods'   # correct fallback to the configured default
   ```

2. *Agent-side stopgap.* Drop the `supports_namespace("database", ...)` requirement and keep only `not supports_namespace("schema", ...)`. Verified this does fix the dotted-name case (`database='dwd'`), and the matrix in my summary shows it is still a no-op for every other shipped adapter. But note it does **not** fix an explicit `database="dwd"` argument: with `capabilities=set()`, `_determine_field_order()` returns `['table']`, so `_default_field_value("database", ...)` returns `""` and the caller-supplied database is silently dropped. Only option 1 covers that.

datus/tools/func_tool/database.py
```diff
@@ -889,10 +910,17 @@ def _build_table_coordinate(
                 setattr(coordinate, field, self._normalize_identifier_part(parsed[parsed_field]))
         return coordinate
 
-    def _table_matches_scope(self, coordinate: TableCoordinate) -> bool:
+    def _table_matches_scope(self, coordinate: TableCoordinate, connector: Optional[BaseSqlConnector] = None) -> bool:
```
stukid, line 913: **Routing is only applied to the table-level check — `list_databases` / `list_schemas` still use primary-connector patterns.**

`_database_matches_scope` and `_schema_matches_scope` still read `self._scoped_patterns` (parsed against `self._primary_connector`) and `self._field_order` (derived from the primary connector), so they do not follow the routed datasource the way `_table_matches_scope` now does.

Verified on this branch:

```
primary connector = postgres, scope = "dwd.*", routed datasource = mysql
  pattern parsed under primary(pg):  database='' schema='dwd'
  list_tables    -> 'other.events' in scope?  False   # correctly rejected
  list_databases -> 'other'        in scope?  True    # slips through
```

`_matches_catalog_database` sees an empty `pattern.database`, sets `wildcard_allowed = True`, and every database passes. A scoped sub-agent can therefore still enumerate out-of-scope database names via `list_databases`, which is the exact hole the routing commit was meant to close.

Suggest either threading the routed connector into these two matchers as well, or documenting in the PR description that they are intentionally out of scope for this change.

datus/tools/func_tool/database.py
```diff
@@ -889,10 +910,17 @@ def _build_table_coordinate(
...
         if not self._scoped_patterns:
             return True
-        return any(pattern.matches(coordinate) for pattern in self._scoped_patterns)
+        patterns = self._scoped_patterns
+        if connector is not None:
+            patterns = []
+            for pattern in self._scoped_patterns:
+                routed_pattern = self._parse_scope_token(pattern.raw, connector)
```
stukid, line 920: Two smaller things about this loop.

**1. Hot-path re-parsing — measured 19x slower per call.**

`_filter_table_entries` calls `_table_matches_scope` once per table entry, and every production call site (`_filter_table_entries`, `_check_sql_table_scope`, `describe_table`, `get_table_ddl`) passes a non-`None` connector. So the `self._scoped_patterns` cache is effectively dead for table-level checks and every entry re-runs `parse_table_name_parts` for every pattern.

Measured on this branch with 5 scope patterns:

```
cached patterns :   0.93 us/call
re-parsed       :  17.99 us/call   (19x)
```

Not fatal, but it scales with `entries x patterns` on `list_tables`. A small memo keyed by dialect + effective capabilities would restore the old cost.

**2. Silent fail-closed when every pattern drops.**

If `_parse_scope_token` returns `None` for all patterns under some routed connector (e.g. an adapter `get_identifier_parser` that yields an empty `table_name`), `patterns` ends up empty and `any([])` is `False` — every table is reported out of scope, with no log line to explain it. Worth falling back to the original pattern and emitting a `logger.warning` instead of dropping silently.

tests/unit_tests/tools/func_tool/test_database.py
```diff
@@ -107,6 +107,110 @@ def test_table_semantic_profile_store_disabled_when_size_probe_fails(self):
...
+        assert pattern.database == "dwd"
+        assert pattern.schema == ""
+        assert tool._table_matches_scope(
+            tool._build_table_coordinate("dwd.events", connector=connector), connector=connector
+        )
+        assert not tool._table_matches_scope(
+            tool._build_table_coordinate("other.events", connector=connector), connector=connector
+        )
+
+    def test_scope_uses_routed_connector_namespaces(self) -> None:
```
stukid, line 189: **This test does not actually pin the routing behavior — it only pins the new signature.**

On the pre-fix code, removing just the `connector=routed_connector` kwarg makes the same assertion pass:

```python
# run against 456fe630 (pre-fix)
tool._primary_connector = SchemaConnector()
tool._field_order = ["schema", "table"]
tool._scoped_patterns = [tool._parse_scope_token("dwd.*")]
coordinate = tool._build_table_coordinate("other.events", connector=DatabaseConnector())
assert not tool._table_matches_scope(coordinate)   # passes on pre-fix
```

Both connectors fall back to the widest field order in the test environment, so `dwd.*` -> `schema='dwd'` and `other.events` -> `schema='other'` mismatch regardless of which connector resolves the pattern. The test is red on the base commit only because of `TypeError: unexpected keyword argument 'connector'`.

So commit `49cc856` currently has no behavioral coverage. To pin it you need a case where primary and routed connectors disagree **and the verdict flips** — for example the postgres-primary / mysql-routed setup from my other comment, where the pre-fix code lets `other.events` through as a wildcard match and the fixed code rejects it.

tests/unit_tests/tools/func_tool/test_database.py
```diff
@@ -107,6 +107,110 @@ def test_table_semantic_profile_store_disabled_when_size_probe_fails(self):
         assert tool._table_semantic_profiles is None
 
 
+class TestDBFuncToolTableCoordinates:
+    @pytest.mark.parametrize(
```
stukid, line 111: The `{"database"}` + `dialect="hive"` combination in the first parametrized case does not correspond to any shipped adapter — `datus-hive` declares `capabilities=set()` (see my comment on `_normalize_parsed_namespaces`). Because the fake connector hardcodes the capability set, the test passes while the real Hive path stays broken.

Per the repo's bug-fix rule ("assert the property, not the instance"), it would be more durable to derive the capabilities from the registry the way production does — e.g. register the dialect with the capability set the adapter actually ships and let `get_effective_capabilities()` fall through to `connector_registry.get_capabilities(dialect)`, as `BaseSqlConnector` does. That way the test tracks the adapter declaration instead of a hand-written assumption about it.

Minor, related: all four cases carry `dialect = "hive"`, including the schema-only and database+schema cases that are meant to represent Postgres-shaped and Snowflake-shaped connectors. Using the matching dialect per case would make the intent clearer.

stukid (commented):
Reviewed by checking the change against the capability declarations of every shipped adapter, and by running the reported scenario end to end.

The code itself is clean and the direction is right, but I believe **this does not fix #1317**. The guard `supports_namespace("database", ...)` is never true for a real Hive connection, so `_normalize_parsed_namespaces` never fires on the Hive path (details inline). The one configuration whose behavior actually changes is MaxCompute with schema disabled — a genuine fix, but not the one in the title.

I ran a matrix of `parse_table_name_parts("a.b", dialect)` before/after the change, using each adapter's registered capabilities:

```
dialect                a.b (pre-fix)     a.b (this PR)     changed
hive                   db='' sch='a'     db='' sch='a'
mysql/spark/clickhouse db='a' sch=''     db='a' sch=''
doris/starrocks        db='a' sch=''     db='a' sch=''
trino/oracle           db='' sch='a'     db='' sch='a'
postgresql/snowflake   db='' sch='a'     db='' sch='a'
maxcompute             db='' sch='a'     db='' sch='a'
maxcompute (no schema) db='' sch='a'     db='a' sch=''     <-- only row that changes
```

So no regression risk for any shipped adapter — the concern is only that the target case is untouched.

Local verification (all green, matches what the description claims): `tests/unit_tests/tools/func_tool/test_database.py` 167 passed · `tests/integration/tools/test_mcp_server.py` 52 passed · `tests/integration/tools/test_func_tools_db.py` 41 passed · `ruff format --check` + `ruff check` clean · `ci/audit_tests.py --diff-only upstream/main` P0=0 P1=0.

Suggested path: land the adapter-side capability fix first, then keep this PR but retitle it to what it actually fixes (MaxCompute-without-schema + scoped-pattern routing), and add a Hive regression test built from the real registry declaration.

simonyang08:
Thanks for the detailed review. I reproduced the capability issue against the actual Datus-ai/datus-db-adapters source and have converted this PR back to Draft.

`datus-hive` 0.1.3 currently registers `capabilities=set()`. Using that production registration with this branch:
- `dwd.orders` becomes `schema=dwd`;
- an explicit `database="dwd"` is dropped;
- a bare table does not recover the configured `ods` database.

A separate local adapter patch changing only the registration to `capabilities={"database"}` makes those three cases resolve as `database=dwd`, `database=dwd`, and `database=ods`, respectively. Its registry contract test fails on the adapter base and passes with the patch; all 45 datus-hive unit tests pass.

I will keep that adapter root fix separate. I will not expand or re-promote #1324 while the remaining routing, cache/failure-semantics, production-metadata fixture, and behavioral-red-test points are unresolved. After the adapter fix is settled, I will reevaluate whether the generic routing changes remain independently valuable and either narrow/retitle this PR or close it rather than continue claiming it fixes #1317.

simonyang08:
Closing this Draft because the current implementation does not fix the reported Hive path. Review and reproduction showed that the behavior depends on adapter capability metadata, while this PR only changes generic routed-scope matching; the branch is also conflicting with current main. The adapter-side root-cause work can proceed separately.

PR #1329, opened 2026-08-24
[Feature] Expose uploaded data files as SQL tables via load_file_as_table

arnozhang:
## Why

A user drops `jeffshop.xlsx` into a project and asks the agent to summarise Q3 sales. Today that fails: `read_file` decodes as UTF-8 and returns `Cannot read binary file`, which tells the model the read failed but not what to do instead — in practice it then burns turns trying `bash cat`. Even for CSV, `read_file` caps at 200 KB and inlining a large file into the turn is the wrong shape for analysis work anyway; aggregation, joins and charting all want SQL.

Nothing in the agent could turn a data file into something queryable.

## Solution

`load_file_as_table` registers a file as SQL table(s) in a DuckDB catalog exposed as a separate datasource (`local_files`), so the whole existing analysis surface — `execute_sql`, `list_tables`, `describe_table`, charts, artifacts — works on it unchanged. `execute_sql` already accepted a `datasource` argument, so no tool signature changed and the default route is byte-identical to before.

**Lazy views, not copies.** Each unit (a sheet, or the whole file) becomes a `VIEW` over the path. Nothing is copied, so there is no second source of truth to keep in sync: re-running a load is idempotent and there is no content hash, registry or refresh command to get wrong. `materialize=True` opts into a snapshot when a large file is queried repeatedly.

How current a view is differs by format, and the docstring says so rather than overclaiming: CSV/Parquet/JSON are fully lazy; spreadsheets are lazy within a pinned range, because `read_xlsx` requires a complete `range` and the end cannot be padded (see below), so rows appended past the recorded end need a reload. Hence the tool telling the model to load before analysing — idempotent and near-free when nothing changed is a cheaper rule than any staleness check.

**Ownership lives in the catalog.** Each object carries a `COMMENT` naming its source file and sheet. Without it a reload cannot distinguish "this name is mine, replace it" from "this belongs to another file, pick another name", and either guess is broken: always replacing clobbers an unrelated upload, always suffixing makes reloads non-idempotent and grows the catalog without bound. The comment doubles as provenance for `describe_table`.

**Table names are ASCII.** `[a-z0-9_]` only, derived from file stem + sheet, with a positional (`_s2`) or hashed (`t_9f3a2b1c`) fallback when nothing ASCII survives — so the model never has to quote a table name. Romanising CJK was rejected: multi-reading characters make it ambiguous, and a plausible-but-wrong name is worse than an honest opaque one. The readable mapping travels in the tool result. Column names keep their originals, since a column's meaning cannot survive transliteration and the model needs it.

**Three findings that shaped the implementation**, all verified against DuckDB 1.5.2:

1. Padding a spreadsheet `range` past the data is *silently wrong*, not merely wasteful. Every blank cell reads back as a NULL row: a 4-row sheet read as `A3:D200` reports 197 rows, a 98% null rate and a phantom NULL group in every aggregate — indistinguishable from genuinely sparse data. `range='A3'` alone is rejected, so an exact end bound is mandatory, which is why openpyxl (not DuckDB) supplies `max_row`/`max_column`.
2. DuckDB cannot enumerate sheets — no `xlsx_sheets()`, and `read_xlsx(..., sheet='nope')` reports "not found" without listing what exists. openpyxl is therefore a hard dependency of the main path, not just the `.xls` fallback.
3. The default read of a workbook whose header sits below a title and a blank row yields the title as the sole column name and **zero rows** — the model concludes the file is empty. Header detection probes with `all_varchar=true, header=false, stop_at_empty=false` (the last is mandatory; it defaults on and truncates the probe at the spacer row), and `header_row` / `inspect_only` let the model correct a bad guess.

**Format coverage.** `.xlsx`/`.xlsm` via the `excel` extension; `.csv`/`.tsv`/`.parquet`/`.json`/`.jsonl` via statically-linked core readers. `.xls` is unsupported by DuckDB entirely (`read_xlsx` is strictly OOXML/zip), so it converts once to Parquet via pandas+xlrd and joins the Parquet path — the only format needing an explicit staleness check, being the only one not read lazily. `.xlsb`/`.ods` return a "re-save as .xlsx" error rather than a half-working path.

**Security.** `load_file_as_table` classifies the path through the same `fs_path_policy` zones as `read_file` (INTERNAL/WHITELIST only) before embedding it in a view. Separately, `execute_sql` refuses file-reading table functions (`read_csv_auto`, `read_parquet`, `glob`, `read_text`, …) **on the uploads datasource only**: that datasource has `enable_external_access` on so lazy views work, and a hand-written `read_csv_auto('/etc/passwd')` would use the same capability to route around the filesystem path policy — the agent's containment boundary. The scope is deliberately narrow: a project that configures its own DuckDB datasource already had that reach, and narrowing it here would be an unrelated behaviour change.

`load_file_as_table` is ALLOW in every permission profile. It reads a file already inside the project and defines a view in a private per-project catalog; falling through to `default=ASK` would prompt for "open the file the user just asked about", and on the API surface — which has no interactive broker — an ASK is effectively a refusal.

Supporting changes: `DuckdbConnector.exclusive_connection()` yields the raw connection under the connector's lock, since a multi-statement unit needs cursor metadata and must hold the lock across all of it (and opening a second `duckdb.connect()` on the same file trips DuckDB's instance-config equality check). The connector also honours `DATUS_DUCKDB_EXTENSION_DIRECTORY`: DuckDB has no env-var equivalent for `extension_directory` — verified, it silently falls back to `$HOME/.duckdb` — and a container's build user and runtime user need not share a HOME, so a deployment that bakes the `excel` extension into its image has no other way to point at it.

## Test Cases

Unit — `tests/unit_tests/tools/db_tools/test_data_file_loader.py` (new, 74 cases): naming determinism and collision stability, identifier legality across CJK/space/punctuation inputs, the SQL guard (including readers nested in subqueries and CTEs), header detection against the title/blank/header shape, CSV/TSV/JSON/JSONL/Parquet loading, lazy-view freshness, materialize round-trips in both directions, idempotency and cross-file name collisions, ownership comments, and every unsupported-format error path. Scoped to what needs no DuckDB extension so it stays deterministic in CI.

Integration — `tests/integration/tools/db_tools/test_data_file_loader_excel.py` (new, 35 cases, `@pytest.mark.acceptance`, registered in `PR_ACCEPTANCE_TARGETS`): sheet discovery, empty-sheet skipping, header/range correctness including a regression test pinning that no all-NULL padding rows appear, freshness within and beyond the pinned range, idempotency, CJK sheet/column handling, `.xls` → Parquet conversion and cache invalidation, and `inspect_only`. Extension-dependent, so it lives here rather than in the unit tier; the fixture `LOAD`s first (the production posture with a baked extension) and only falls back to `INSTALL`.

Extended — `test_database.py` (uploads-catalog SQL guard, including that the connector is never touched on rejection, and that other datasources are unaffected), `test_filesystem_tools.py` (`read_file` redirect, case-insensitivity, CSV still reading normally, missing files still reporting not-found), `test_profiles.py` (ALLOW in all three profiles without widening other `db_tools`), `test_func_tools_db.py` (tool-surface contract updated for the new tool).

`xlwt` added to the dev group so the legacy `.xls` happy path runs in CI rather than skipping.

Gates: `ruff format` / `ruff check` clean. `ci/audit_tests.py --all` reports **P0=0** with zero findings of any severity in the new/changed test files. Full unit suite 18182 passed; the acceptance suite goes from 65 to 94 passed with the same 17 pre-existing failures — verified by diffing failure lists against `upstream/main` with the changes stashed, so zero regressions.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
- [ ] release/0.3.8
- [ ] release/0.3.9
- [x] release/0.4.0
## Summary by CodeRabbit

- **New Features**
  - Added support for loading CSV, TSV, JSON, Parquet, and Excel files as queryable data.
  - Added file inspection, previews, row counts, column summaries, sheet selection, and optional table materialization.
  - Added legacy `.xls` conversion and caching.
  - Added agent guidance for querying uploaded files.

- **Bug Fixes**
  - Prevented unsupported direct file-reading SQL on the local-files datasource.
  - Added clearer handling for unsupported binary file formats and spreadsheet errors.

- **Tests**
  - Added comprehensive coverage for file loading, Excel support, permissions, and access protections.

## Summary by CodeRabbit

- **New Features**
  - Added robust CSV/TSV encoding detection for UTF-8, UTF-16, and GB18030 files.
  - Added support for loading and selecting individual sheets from legacy `.xls` workbooks.
  - Uploaded files can be registered and queried through the local files data source.
  - Added semantic model selection and richer table and column descriptions.
  - Improved workspace-aware file path handling.

- **Bug Fixes**
  - Spreadsheet imports now ignore formatting-only cells and preserve data after blank rows.
  - Improved handling of invalid sheets, ranges, encodings, and empty workbooks.
  - Added clearer guidance for unsupported `.xlsb` and `.ods` files.
  - Strengthened safeguards for unsafe or unregistered uploaded-file SQL access.


arnozhang:
## Review addressed — all 11 findings

Both blocking findings reproduced exactly as described, and one of them was worse than the report suggested. Details below; the fixes are in `ce0278d4` (agent) / `9d345ed` (backend) / `e2860a12` (saas).

### 🔴 1 — `sheet_bounds` reintroduced the padding bug it warned about

Confirmed, and I took the second of your two options. A minimal repro:

```
sheet: 2 data rows; one bolded blank cell at B200, one italic blank at E5
openpyxl extent      -> (200, 5)
used_range           -> A1:E200
row_count            -> 199          (real: 2)
null_percentage      -> [98.99, 98.99, 100.0, 100.0, 100.0]
columns              -> ['region', 'qty', 'C2', '_1', '_2']
GROUP BY region      -> [('east',1), ('west',1), (None,197)]
```

`stop_at_empty=true` was tempting for the one-line fix, but interior blank rows are how hand-maintained exports separate groups, and truncating there loses real data with no signal. So `sheet_bounds` now scans for the last cell holding a value and `_xlsx_scan` states `stop_at_empty=false` explicitly — the exact range means there is no padding for it to protect against, so the two changes are complements rather than alternatives. After: `A1:B3`, 2 rows, no phantom group or columns; and a separator-row fixture keeps all three groups.

### 🔴 2 — gitignore blinded the agent's only discovery tool

Confirmed, and worse than "returns empty": it breaks the *first step* of the feature's headline scenario. `glob("**/*.xlsx")` → `{'files': []}` with the file sitting right there.

Took your first option. The rules move to `.git/info/exclude`: git honours it identically, `_load_gitignore_patterns` only ever looks for `.gitignore`, and it is untracked so it also stops appearing in the user's diffs. `.gitignore` keeps only the heartbeat rule, so that behaviour is unchanged. Verified in the running service:

```
.gitignore            -> ['.datus-heartbeat-*']
.git/info/exclude     -> ['*.xlsx', '*.xlsm', '*.xls', '*.parquet']
glob('**/*.xlsx')     -> {'files': ['jeffshop_q3_2017.xlsx']}
git status --porcelain -> (no .xlsx)
git check-ignore -v   -> .git/info/exclude:7:*.xlsx
```

I did not take the "make the fs tools skip datus-injected rules" option — that couples the filesystem policy to git housekeeping, and the coupling would have to be re-remembered every time either side changes.

### 🟡 3 — root-owned extension directory

Confirmed by inspection: `chown -R` covers `/var/log/datus /data/tenants /app/.cache`, and the env var redirects *every* extension install, so iceberg's `INSTALL httpfs` would hit permission denied after `USER datus`. Added `/app/.duckdb` to the chown — chosen over baking the rest because the set of extensions a deployment might reach for is open-ended.

### 🟡 4 — unreachable write ceiling

Confirmed: no `ws_max_size` anywhere, so 16 MiB default against a 50 MB decoded limit. `ws_max_size` is now derived from `MAX_WRITE_FILE_BYTES` (× 4/3 for base64, plus envelope headroom) and overridable via `UVICORN_WS_MAX_SIZE`, so the two limits cannot drift apart again and an oversized write returns `FileTooLargeError` instead of closing the socket.

### 🟡 5 — `.csv` still corrupted

You are right, and the PR text you quoted was the wrong argument: `read_file` being able to read CSV says nothing about what the *upload* path should do. I went further than adding `.csv` to the list — the list is gone. Every upload is `ArrayBuffer` + base64 now. Any "which formats are text" list is a correctness dependency whose failure mode is silent and unrecoverable, and the only thing it bought was 33% of wire size on text files, which finding 4's frame ceiling now covers.

### 🟢 6–11

| # | Resolution |
|---|---|
| 6 | `load_file_as_table` refuses when the uploads datasource is absent, so legacy single-connector mode can no longer land `CREATE VIEW` in the user's real database |
| 7 | Added `filesystem_root` to `DBFuncTool`, passed from the chat node's `_resolve_workspace_root()`; defaults to `project_root`, which is the same value unless a node overrides it |
| 8 | `(str(exc).splitlines() or [repr(exc)])[0]` — the handler no longer raises from inside the "don't sink the workbook" path |
| 9 | Promoted past LOW: `ATTACH` / `COPY TO` are not function calls, and the tenant PVC is mounted. Rather than extend the blacklist, `execute_sql` on that datasource now allows **only** SELECT / SHOW / DESCRIBE / EXPLAIN, with unparseable SQL failing closed. Nothing legitimate is lost — every write comes from `load_file_as_table`'s own connection |
| 10 | `.xlsb` / `.ods` answer in one turn ("re-save as .xlsx or .csv") instead of redirecting to a tool that also rejects them |
| 11 | `local_files` filtered out of `_collect_metadata`'s datasource list |

### Also found while verifying

`available_tools()` is hand-curated, not derived from `all_tools_name()` — so `load_file_as_table` was permissioned, catalogued, and **never mounted for the model**. The feature was dead in the real agent, and every unit test passed because they all asserted the catalog surface. Fixed, plus tests that pin the LLM-facing list specifically (mounted with the catalog, absent without it, and single-connector mode still building its list at all).

### Tests and gates

New regression coverage for the two blocking findings — styled-blank extent (range / row count / phantom group / phantom columns), interior blank row preserved, `glob` visibility end-to-end over both git mechanisms, non-read statement rejection incl. `ATTACH` and `COPY TO`, unparseable-SQL fail-closed, and the LLM-surface mount tests.

Backend **1255 passed / 39 skipped**. Agent: failure set byte-identical to `upstream/main` (56 pre-existing, 0 new), verified by stashing and re-running. `ruff` clean; `ci/audit_tests.py --all` **P0=0** with zero findings of any severity in the new/changed test files.

arnozhang:
## Round 2 — all five addressed

`9aeff065` (agent) / `8a3d590` (backend) / `7f215ff1` (saas).

### 🔴 A — PR 855 had 9 unrelated files

Mine, and a bad one: I used `git add -A` in a repo where in-progress SQL-autocomplete work was sitting unstaged, and swept it into my commit. The saas tree was clean when I branched, so this arrived mid-run.

Reset the commit and unstaged those 9 paths rather than reverting them, so the work is back in the working tree untouched — nothing lost. PR 855 is now **3 files** (`upload-utils.ts`, the fs store, the WS payload types). Force-pushed with `--force-with-lease`.

Agreed on your read of that change, and agreed it wants its own PR: no reason for a prefetch fix to inherit 855's "wait for backend#420" ordering constraint.

### 🟡 B — replacement scan

Confirmed, and it's the sharpest thing in either review round. Reproduced on 1.5.2:

```
SELECT * FROM '<path>/other_tenant.csv'   -> stmt=select  guard_fns=[]  READ [('root','hunter2')]
SELECT * FROM '<path>/*.csv'              -> stmt=select  guard_fns=[]  READ [('root','hunter2')]
```

Took your second suggestion — the whitelist — because the first still leaves a blacklist to keep complete. Every table reference must now resolve to an object the catalog holds, or to a CTE in the same statement.

One thing worth flagging from implementing it: sqlglot lowers a quoted path into an `Identifier`, so `'/tmp/x.csv'` and `sales` are the *same node type*. Discriminating on `Literal` (your parenthetical) wouldn't have fired — the only difference is that a path isn't in the catalog, which is exactly what resolution tests. And table *functions* carry their name on the inner node rather than on the `Table`, so a naive walk reports them as unnamed and skips them — i.e. it would exempt precisely the constructs this exists to catch, including whatever reader DuckDB adds next. Both are covered.

I kept the function-name gate in front of it, which your report implied was now redundant. It isn't quite: it catches a reader called where a *value* goes rather than where a table goes, which leaves no table reference behind at all. `SELECT read_text('/etc/hosts')` is the case — DuckDB happens to reject that one itself ("table function used as a scalar"), so it's belt-and-braces rather than a known hole, but it costs one regex and gives a better message for the common `FROM read_csv_auto(...)` shape.

After:

```
SELECT * FROM '<path>/ok.csv'          BLOCKED
SELECT * FROM '<path>/**/*.csv'        BLOCKED
SELECT * FROM read_csv_auto('/etc/…')  BLOCKED   read_csv_auto() is not a table registered
ATTACH '/etc/passwd' AS x              BLOCKED
COPY (SELECT 1) TO '/tmp/x.csv'        BLOCKED
SELECT * FROM ok                       ALLOWED   (registered)
```

Failing to read the catalog now refuses rather than skipping the gate.

### 🟡 C — retired rules linger

Right, and it's the worse half of the bug: the fix landed for new workspaces and left it in place exactly where the branch had already been deployed. The managed rules are reconciled in both directions now — retired dropped, current appended, and a rule the user added themselves survives (asserted).

### 🟢 D — `filesystem_root` only on chat

Confirmed: 19 sites across 11 files, and `available_tools()` mounts the tool wherever the datasource exists, so the rest were re-anchoring.

I first did this as a shared `AgenticNode._make_db_tool` factory — tidier, and a 20th site couldn't forget. It broke three tests, all for the same reason: several node tests patch `<module>.DBFuncTool` as their seam, and routing construction elsewhere removes that attribute. Not worth breaking an established seam for an aesthetic, so I reverted to your shape — one line per site — and moved the "can't forget" guarantee into a test that walks all 19 argument lists. Enforced beats structural here anyway.

### 🟢 E — vscode anchor

Agreed it isn't a regression, and agreed the honest answer is to say so: `load_file_as_table` now refuses on a vscode session and explains that the workspace lives on the client, instead of resolving against the daemon CWD and reporting an inexplicable "File not found".

### Gates

Agent: failure set byte-identical to `upstream/main` (56 pre-existing, 0 new), verified by stashing and re-running. Backend **1256 passed / 39 skipped**. `ruff` clean both repos; `ci/audit_tests.py --all` **P0=0**, zero findings of any severity in the new/changed test files. New regression coverage for B (paths, globs, joins, CTE-wrapped, table functions incl. an invented future one, qualified names, case-insensitivity, unparseable → fail closed), C (legacy workspace prune, user rule preserved, glob visibility, git still ignoring), D (all-sites assertion + explicit-root precedence), E (refusal message, web unaffected).

arnozhang:
## CodeRabbit review — verified each against current code

Several findings describe the pre-round-2 state; those I've closed with evidence rather than a change. Fixes in `e67ba8c1`.

### Fixed — two were real bugs, and one was mine to answer for

**`sheet_bounds` dropped trailing formula rows.** The worst of the batch, and it came from my own round-2 fix. `data_only=True` yields a formula cell's *cached* value, and a workbook written by a library rather than by Excel has no cached values at all — so the cell reads `None`, its row falls outside the bounds, and the data is gone silently. Exactly the mirror of the padding bug the function exists to prevent:

```
sheet: header, 2 data rows, A4 = "=A2"  (no cached value)
before -> bounds (2, 2)   range A1:B2   the formula row dropped
after  -> bounds (4, 2)   range A1:B4   3 rows
```

Bounds are now read with formulas visible. Regression tests cover a trailing formula row and a formula-only column.

**`.xls` ignored `sheet`.** `pandas.read_excel` defaults to the first worksheet, so asking for the second sheet of a multi-sheet workbook returned the first one's data — wrong answers rather than an error, which is the worse failure:

```
before -> sheet='Second' gave cols ['a'] rows [(1,)]      # that is First
after  -> sheet='Second' gave cols ['b'] rows [(99,)]
          sheet=None     gave multi_first + multi_second
```

Legacy workbooks now behave like `.xlsx`: one table per sheet, one cached conversion per sheet (keying the cache on the file alone would have had the second sheet overwrite the first — asserted).

**`header_row` validation.** It did already fail rather than corrupt, but by leaking `Binder Error: Invalid range 'A0:B2'`. Since the model supplies this value, it now gets `header_row must be 1 or greater, got 0`.

**`exclusive_connection` lock ordering.** Correct as reported — `connect()` then `with self._lock` leaves a window where a concurrent `close()` lands between them and the caller is handed `None`. Lock is now held across both; `connect()` takes the same RLock re-entrantly.

**`read_only` on `load_file_as_table`.** Agreed — it creates catalog objects, so it belongs with the other write entry points. `inspect_only` creates nothing and stays available.

**`extension_directory` through `_sql_literal`**, and **`default_conversion_cache_dir("")`** no longer resolves into the process working directory (temp dir instead).

**CI network.** This one I'm glad you raised: my integration fixture could reach `extensions.duckdb.org`, which the repo's own guardrail forbids for CI tests. The workflow now installs the extension once up front and points `DATUS_DUCKDB_EXTENSION_DIRECTORY` at it, so the suite loads locally — same posture as the runtime image. The `INSTALL` fallback stays for developer machines and is deliberately *not* wrapped in a skip: an environment that can neither load nor fetch it should fail loudly, since that is the breakage most likely in production.

**`detect_header_row` docstring** corrected — it claimed the following row must have equal or greater width; the rule is simply that it also has at least two non-empty cells.

**Test hardening** — case variations, whitespace before the paren, an allowed near-miss (`SELECT read_csv_auto FROM …`, a column merely *named* like a reader), and the allowed paths now assert `execute_query` / `execute_ddl` were actually called rather than only that no rejection message appeared.

### Skipped, with evidence

**Add `sniff_csv` / `read_json_objects` / `parquet_metadata` / … to `FILE_READING_FUNCTIONS`, and make the guard see replacement scans.** Already closed, in the previous round: the table-reference gate is a whitelist now, so every one of those is rejected without being named anywhere:

```
sniff_csv             -> ['sniff_csv()']
read_json_objects     -> ['read_json_objects()']
read_ndjson_objects   -> ['read_ndjson_objects()']
parquet_metadata      -> ['parquet_metadata()']
parquet_file_metadata -> ['parquet_file_metadata()']
parquet_schema        -> ['parquet_schema()']
```

That is the point of a whitelist — a name list would need extending again for the next reader. Replacement scans are likewise covered (tests assert absolute paths, globs, joins and CTE-wrapped forms).

**Lazy view + symlink escape.** `classify_path` resolves symlinks, so the escape is rejected at validation, not at query time — verified: a link inside the project pointing outside classifies `EXTERNAL`. The residual TOCTOU needs an attacker who can already write into the project, and the proposed remedy (force materialisation for externally mutable sources) would trade away the feature's core property — that a query always sees the file's current contents — for a threat that write access would achieve more directly.

**`.xlsb`/`.ods` should not redirect to `load_file_as_table`.** Already fixed a round earlier: `_UNREADABLE_TABULAR_EXTENSIONS` is separate and answers "re-save as .xlsx or .csv" in one turn.

**`DataFileError` → `DatusException`.** It never crosses the module boundary — `load_file_as_table` catches it and returns a `FuncToolResult`. That matches existing in-tree precedent for internal control flow (`PermissionDeniedException`, `AttributionValidationException`, `BundleError`, all plain `Exception`). Converting it would add an error code for something no caller can observe as an exception.

**Move `PREVIEW_ROW_LIMIT` / the probe limits to `agent.yml`.** These are internal output-budget constants, not deployment knobs; `FilesystemConfig.max_file_size` sets the precedent for that class of limit staying a constant. Three YAML keys nobody has asked to tune is surface without a use.

**`extension_directory` as a YAML datasource field.** Scope creep for this PR — it is a per-deployment fact about where the image put the binary, not a per-datasource one, which is why it is an env var. Happy to add it separately if operators want per-datasource control.

**Open each workbook once.** Real inefficiency, and mine, but bounded: the extra passes are over a file every query already re-reads. Left out to keep this diff to correctness; worth doing on its own if spreadsheet loading shows up in a profile.

**`requirements-test.txt` GitHub archive URL.** Pre-existing — that pin is not from this PR; I only regenerated the file via `uv export` when adding `xlwt`.

### Gates

Agent failure set byte-identical to `upstream/main` (56 pre-existing, 0 new), verified by stashing and re-running. Backend 1256 passed / 39 skipped. `ruff` clean; `ci/audit_tests.py --all` **P0=0**, zero findings of any severity in the new/changed test files.

arnozhang:
## CSV: supported all along, but only for UTF-8 — now actually supported

`.csv` / `.tsv` have been first-class since the first commit (`read_csv_auto`, tests, and the reason the IDE upload switched to raw bytes). Asked to confirm it, I tested the case the byte-preservation change was *for* and it failed:

```
utf8.csv  -> cols ['门店','品类','金额']  rows OK
gb.csv    -> Invalid Input Error: CSV Error on Line: 1
             Invalid unicode (byte sequence mismatch) detected. This file is not utf-8 encoded.
```

DuckDB's reader assumes UTF-8 and fails the whole read otherwise. So the saas-side argument in #855 — *let DuckDB see the real bytes, that's more correct* — got the bytes right and stopped one step short: they arrived at a reader that could not decode them. Excel's "Save as CSV" on a Chinese Windows writes GB18030, which makes this the common case, not an edge one.

DuckDB 1.5.2 does accept `encoding`, so the fix is to work one out and pass it. Fixed in `9610143e`.

### What it does now

```
utf8.csv          enc=utf-8      cols ['门店','品类','金额']
bom.csv           enc=utf-8      (BOM decides)
gb.csv            enc=gb18030    cols ['门店','品类','金额']
gb.tsv            enc=gb18030    tab delimiter preserved
u16.csv           enc=utf-16
shift_jis.csv     enc=shift_jis  cols ['店舗','金額']
cp1252.csv        enc=cp1252     cols ['café','montant']
big5.csv          REFUSED with the remedy — see below
```

### Two things the implementation had to account for

**The detected label is not trustworthy on its own.** `charset_normalizer` (already a runtime dependency, so no new one) labelled my GB18030 sample `cp949`, the Shift-JIS one `cp932`, and the cp1252 one `cp1250`. On a short sample the CJK double-byte ranges overlap that much. Passing the label straight through would fail the read for a file it had just correctly identified as non-UTF-8. So the label only *orders* the candidates and the choice is confirmed by actually decoding — incrementally, since a fixed-size sample almost always ends mid-character and a one-shot decode would reject the right encoding.

**Big5 is accepted by DuckDB and then quietly corrupted.** Worth flagging as a DuckDB limitation rather than something I worked around:

```
bytes                b'\xaa\xf9\xa9\xb1,\xaa\xf7\xc3B\n\xa5x\xa5_,299.5\n'
python .decode(big5) '門店,金額\n台北,299.5\n'                      # correct
duckdb encoding=big5 cols ['門店','金額◙台北','299.5◙'] rows []     # row boundary lost
```

A Big5 trailing byte can land in the ASCII range (`0x42` here) and the reader loses the newline after it. So Big5 is refused with the remedy ("re-save as UTF-8, or pass `encoding='big5'` to read it anyway") instead of returning rows nobody can tell are wrong. The refusal is gated on the *label*, because gb18030 decodes Big5 bytes without complaint and a decodability-first order lets it shadow the check.

The candidate list is therefore restricted to encodings I verified DuckDB round-trips intact — gb18030, shift_jis, cp1252, plus utf-8/utf-16 from the BOM and short-circuit paths. `latin-1` is deliberately excluded: it decodes any byte sequence, so offering it would mask every real answer behind mojibake.

### Visible and overridable

The chosen encoding is reported on every loaded table and by `inspect_only`, and `load_file_as_table(encoding=...)` overrides it. CJK discrimination is a heuristic; one that cannot be seen cannot be corrected.

15 new tests: per-encoding detection, BOM, ASCII, loader round-trips, a GB18030 `.tsv` keeping its tab delimiter, the reported encoding, the explicit override, the Big5 refusal, latin-1 never being auto-selected, and a truncated sample not defeating detection.

Agent failure set unchanged against `upstream/main` (56 pre-existing, 0 new); backend 1256 passed; `ruff` clean; audit P0=0.

arnozhang:
## Conflicts resolved — merged `upstream/main` (`650148f0`)

Merged rather than rebased so the commit SHAs referenced in the review threads above stay resolvable.

Both conflicts were mechanical:

- **`datus/tools/func_tool/database.py`** — the typing import line only. #1327 added `Set`, this branch added `Path`; kept both. Nothing else in that file conflicted despite #1327 reworking 332 lines of it.
- **`requirements-test.txt`** — generated by `uv export`, so regenerated from the merged lock rather than hand-merged. `xlwt` (dev) and `openpyxl` / `xlrd` (runtime) all survived.

**One thing the auto-merge could not catch.** #1327 renamed `SemanticModelRAG` → `SemanticDatasetRAG`. Git updated the call sites it already knew about, but the five `patch(...)` targets this branch added still named the old class — which fails at patch time, not at import, so it would have shown up as five puzzling test failures rather than an ImportError. Updated.

Verified in the same working directory, before and after, so the accumulated test-data state is identical: **`upstream/main` 56 failures, this branch 56, set-identical, zero new.** (A first attempt used a separate worktree and produced 35 vs 56 with disjoint failure sets — the worktree lacks the accumulated fixture state, so that comparison was meaningless. Noting it in case anyone else reaches for a worktree to baseline this suite.)

`ruff` clean, `ci/audit_tests.py --all` P0=0, backend 1256 passed. The backend PR's submodule pointer is bumped to the merge commit.

### One caveat about the CI step in this PR

This repo's coverage workflow triggers on `pull_request_target`, which by design runs the workflow definition from **base**, not from the PR. So the "Pre-install the DuckDB excel extension" step added here does not execute on this PR's own runs — today's green run reached the network to `INSTALL excel`, which is exactly what the step exists to stop. It takes effect once merged.

Since it can't be demonstrated by CI pre-merge, I ran the step's own commands locally against a clean extension directory and then the suite in the runtime posture:

```
1) install into a fresh dir      -> duckdb 1.5.2: excel extension ready in <dir>
2) suite with that dir           -> 41 passed        (autoinstall never consulted)
3) negative control: empty dir   -> IO Error: Extension ".../excel.duckdb_extension" not found
```

Step 3 is the part that matters — it proves step 2 was reading the pre-installed binary rather than silently falling back to a cache elsewhere.

arnozhang:
## Valid, and worse than described — fixed in `b677d97c`

Confirmed: the candidate loop decodes incrementally precisely so a sample ending mid-character isn't mistaken for a decode failure, and I left the UTF-8 short-circuit doing a strict one-shot decode. Reproduced across every truncation point of a UTF-8 fixture:

```
cut=1  strict=FAIL  incremental=True  -> detect='gb18030'          # misdetected
cut=2  strict=FAIL                    -> DataFileError: "looks big5-encoded"   # refused outright
```

So it isn't only a misdetection: at some cut points a valid UTF-8 file is **refused with a wrong diagnosis**, which is the worst of the three possible outcomes and the exact failure class this PR has been chasing.

Reachable at the production default, not just with a contrived `sample_bytes`. Any UTF-8 file over 256 KiB whose boundary lands inside a multi-byte character:

```
pad=0  boundary strict_decode=FAIL  -> before: fell through   after: 'utf-8'
```

Fix is the one-liner you describe — the short-circuit now goes through `_decodes`. Verified:

```
368 truncation points of a UTF-8 fixture   -> 0 misdetected (was: several, incl. a false Big5 refusal)
gb18030 / utf-16 / shift_jis fixtures      -> unchanged, still detected correctly
gb18030 truncated at 16/32/64/128 bytes    -> never claimed as UTF-8
```

That last check matters because the tolerant decoder is what makes the short-circuit safe: I checked `_decodes` on a GB18030 sample at every length from 2 bytes up and it is `False` throughout, so the tolerance does not swallow non-UTF-8 bytes.

### One thing the fix surfaced

Writing the "does not loosen" test with a deliberately tiny `sample_bytes=7` tripped the **Big5 refusal** instead: on seven bytes the detector confidently labels almost anything, and a valid GB18030 file was refused as Big5. That refusal exists to prevent silent corruption, but a sample that short has no rows to corrupt, so refusing a valid file is the worse error. It is now bounded by sample size — set just past "too short to hold a CSV row", so a real header plus one row still gets refused when it genuinely is Big5 (asserted both ways).

Two regression tests added: every truncation point of a UTF-8 fixture resolving to UTF-8 (a wrong refusal counted as a failure, not just a wrong value), and the tiny-sample case not raising.

Agent failure set unchanged against `upstream/main` — 56 pre-existing, 0 new, same directory before and after. Backend 1256 passed. `ruff` clean, audit P0=0. Backend submodule pointer bumped.

PR #1332, opened 2026-08-24
[BugFix] Bound the /table/columns prefetch so it cannot starve the datasource

arnozhang:
## Problem

A customer on-prem deployment (Oracle) reported the whole app hanging for about a minute after entering a project. The frontend was firing a catalog-wide column warm-up — six parallel `POST /table/columns` of 200 tables each (fixed on the SaaS side in Datus-ai/Datus-saas#856) — but the reason it took the *service* down rather than just being slow is here.

`get_tables_columns` resolves its tables serially, and each `get_table_schema` takes the single per-connector `_schema_lock` while holding an `asyncio.to_thread` worker:

```python
for full_path in tables:
    detail = self.get_table_schema(full_path)   # N serial round-trips, N lock acquisitions
```

So N tables cost N round-trips. On a slow source that is minutes of held lock and a pinned worker, and **everything else needing table metadata queues behind it** — interactive `/table/detail`, the agent's schema linking during chat. Six such batches in parallel could not go any faster (they serialize on the lock regardless) but each pinned a worker while waiting.

## Fix

Prefetch is a convenience, so it should yield rather than compete. Two bounds on top of the existing size cap:

- **One batch at a time**, and a second **does not queue** — it returns what the cache already holds. Queuing is the part that pins a worker, and a queued batch could not go faster anyway.
- **A time budget** (`api.prefetch_budget_seconds`, default 3s). The batch stops when it is spent.

Both leave tables out of the response, which is exactly the contract a table that fails to resolve already has — so a client re-asks for what it still needs, one table at a time, via `/table/detail`. The matching SaaS change evicts omitted tables so its on-demand path retries them.

**Cache hits are served regardless of either bound**, and a fully-warm batch never takes the gate at all — the common steady-state case stays free.

Worst case for one request goes from unbounded (~50s+ observed on Oracle) to the budget plus one table round-trip — the budget is only checked between tables, and a single `get_table_schema` (including its wait on the connector lock) has no timeout of its own. At the ~0.25s/table observed on Oracle that is ~3.25s. At most one batch touches the source at any moment.

Two smaller things in the same area:
- `max_prefetch_tables` no longer raises out of `int()` on a non-numeric `agent.yml` value; a typo falls back to the default instead of taking the endpoint down.
- Table-identity resolution is factored out of `get_table_schema` into `_resolve_table_identity` / `_cached_columns`, so the batch can test the cache without touching the connector.

## Known limits — this is not blanket rate limiting

Three boundaries a reader should not assume are covered. Each is a deliberate
follow-up, not an oversight:

**A single round-trip is still unbounded.** The budget is only evaluated at the
top of the loop, and `get_table_schema` holds `_schema_lock` across
`connector.get_schema()` with no timeout of its own. If one wide table's
`ALL_TAB_COLUMNS` query takes 60s, the batch holds the lock and a to_thread
worker for those 60s and interactive `/table/detail` queues behind it. This
change moves the ceiling from N×T to budget+T, not to budget. A real fix means
a timeout on `get_schema`, or taking the lock with `acquire(timeout=)`.

**The gate is per-project, so it does not defend across projects.**
`DatasourceService` is cached per project and `DBManager` is not a singleton
(`db_manager.py:193` — `_conn_dict` is an instance field), so M active projects
have M independent gates, locks and connectors. On-prem all projects share one
container, so six IDEs open at once means six parallel batches, each pinning a
worker from the shared `min(32, cpu+4)` pool and all hitting the same Oracle.
Smaller than the original blast radius, not gone. Fixing it means lifting the
gate to module scope, or a registry keyed by datasource URI — keyed, because a
module-level gate would serialize projects on unrelated datasources.

**An over-limit batch is rejected whole, which is inconsistent with the other
two bounds.** A spent budget and a busy gate both return `success=True` with a
partial result; only the size cap returns `INVALID_PARAMETERS`, which the SaaS
http client throws on. So an operator who tunes `max_prefetch_tables` below the
client's fixed 50 gets batches that always fail, and a client that releases the
names on failure re-sends the same doomed request on the next keystroke —
nothing ever warms. `tables[:max_tables]` with the remainder reported as omitted
would make all three bounds the same shape. Not done here because it changes a
documented response for the endpoint and the test that pins it; worth its own
change.

## What this does *not* fix

The real cost is that columns are fetched one table per round-trip. Every dialect could do this in a single metadata query (`ALL_TAB_COLUMNS`, `information_schema.columns`), but `get_schema` is a per-table interface on `BaseSqlConnector` in `datus_db_core`, and the dialect connectors live outside this repo. That is the actual fix and it belongs in a separate change; this PR only stops one caller from monopolising the datasource in the meantime.

## Test

New `TestGetTablesColumnsBounds` — 7 cases: budget exhausted before the first table (asserts the connector is never called), omitted tables still fetchable per-table afterwards, budget spent mid-batch returns the partial result, a second batch returning instead of queuing, cache hits served while the gate is held, gate released for the next batch, gate released when a table raises. Plus one for the non-numeric limit falling back.

- `tests/unit_tests/api/services/test_database_service.py` — 71 passed.
- `tests/unit_tests/api` — 1352 passed; the 10 failures + 2 errors in `test_explorer_service.py` / `test_metric_hooks.py` are pre-existing (identical on a clean `main`).
- `ruff format` + `ruff check` clean.

## Related

- Datus-ai/Datus-saas#856 — removes the catalog-wide warm-up that triggered this and batches the per-document prefetch.
- A Datus-backend submodule bump follows once this merges.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
- [ ] release/0.3.8
- [ ] release/0.3.9
- [x] release/0.4.0
## Summary by CodeRabbit

* **Improvements**
  * Added safer and more predictable database column loading with configurable request limits and time budgets.
  * Cached results are returned immediately without additional throttling.
  * Partial results remain available when processing limits are reached.
  * Individual table details can still be retrieved when batch loading is incomplete.
  * Improved handling of invalid settings, concurrent requests, and temporary failures.

## Summary by CodeRabbit

* **Improvements**
  * Added safer, more predictable database column loading with configurable request limits and time budgets.
  * Cached results are returned immediately without additional throttling.
  * Partial results remain available when processing limits are reached.
  * Individual table details can still be retrieved when batch loading is incomplete.
  * Improved handling of invalid settings, concurrent requests, and temporary failures.
  * Standardized validation for session paging and duration settings, including fractional and infinite values.

### DIFF
--- a/datus/api/services/chat_service.py
+++ b/datus/api/services/chat_service.py
@@ -34,6 +34,7 @@ from datus.api.services.chat_task_manager import (
 from datus.configuration.agent_config import AgentConfig
 from datus.models.session_manager import SessionManager, session_matches_agent
 from datus.schemas.action_history import ActionRole, ActionStatus
+from datus.utils.config_utils import coerce_positive_int
 from datus.utils.exceptions import DatusException, ErrorCode
 from datus.utils.loggings import get_logger
 from datus.utils.time_utils import now_utc_iso
@@ -46,22 +47,6 @@ logger = get_logger(__name__)
 # request path — clients that omit ``limit`` entirely — stays bounded.
 _DEFAULT_SESSION_PAGE_SIZE = 50
 _DEFAULT_MAX_SESSION_PAGE_SIZE = 200
-
-
-def _positive_int(raw: Any, fallback: int) -> int:
-    """Coerce a config value or caller-supplied page size to a positive int.
-
-    Both operator YAML and direct (non-HTTP) callers are untrusted here: a
-    missing key, ``null``, a non-numeric string, or a non-positive number all
-    fall back to *fallback* rather than producing an empty or reversed page.
-    ``OverflowError`` is caught alongside the usual pair because YAML ``.inf``
-    parses to float infinity, which ``int()`` refuses to convert.
-    """
-    try:
-        value = int(raw)
-    except (TypeError, ValueError, OverflowError):
-        return fallback
-    return value if value > 0 else fallback
 
 
 class ChatService:
@@ -159,16 +144,16 @@ class ChatService:
         """
         try:
             api_config = getattr(self.agent_config, "api_config", {}) or {}
-            max_page_size = _positive_int(api_config.get("max_session_page_size"), _DEFAULT_MAX_SESSION_PAGE_SIZE)
+            max_page_size = coerce_positive_int(api_config.get("max_session_page_size"), _DEFAULT_MAX_SESSION_PAGE_SIZE)
             default_page_size = min(
-                _positive_int(api_config.get("default_session_page_size"), _DEFAULT_SESSION_PAGE_SIZE),
+                coerce_positive_int(api_config.get("default_session_page_size"), _DEFAULT_SESSION_PAGE_SIZE),
                 max_page_size,
             )
             # A non-positive ``limit`` is treated as unspecified rather than
             # passed through: the route pins ``ge=1``, but a direct caller
             # passing -1 would otherwise slice ``all_ids[offset:-1]`` and
             # enrich nearly every session — the exact cost this bounds.
-            page_size = min(_positive_int(limit, default_page_size), max_page_size)
+            page_size = min(coerce_positive_int(limit, default_page_size), max_page_size)
             # Likewise the route pins ``ge=0``; clamp again for non-HTTP callers,
             # since a negative offset would silently slice from the tail.
             offset = max(offset, 0)
--- a/datus/api/services/database_service.py
+++ b/datus/api/services/database_service.py
@@ -40,6 +40,7 @@ from datus.storage.semantic_model.artifact_file import (
 )
 from datus.tools.db_tools.capabilities import supports_namespace
 from datus.tools.db_tools.db_manager import DBManager
+from datus.utils.config_utils import coerce_positive_int, coerce_positive_seconds
 from datus.utils.loggings import get_logger
 from datus.utils.sql_utils import parse_table_name_parts
 from datus.utils.text_utils import redact_uri
@@ -61,20 +62,6 @@ _DEFAULT_PREFETCH_BUDGET_SECONDS = 3.0
 def _brief_columns(columns: list[ColumnInfo]) -> list[TableColumnBrief]:
     """Slim the prefetch payload: no default_value, which no client reads."""
     return [TableColumnBrief(name=c.name, type=c.type, nullable=c.nullable, pk=c.pk) for c in columns]
-
-
-def _positive_number(raw: Any, fallback: float) -> float:
-    """Coerce an operator-supplied bound to a positive number.
-
-    Operator YAML is untrusted here: a missing key, ``null``, a non-numeric
-    string, or a non-positive number all fall back rather than raise, since a
-    typo in agent.yml must not take the endpoint down.
-    """
-    try:
-        value = float(raw)
-    except (TypeError, ValueError):
-        return fallback
-    return value if value > 0 else fallback
 
 
 class DatasourceService:
@@ -594,7 +581,7 @@ class DatasourceService:
         what it still needs — one table at a time, via /table/detail.
         """
         api_config = getattr(self.agent_config, "api_config", {}) or {}
-        max_tables = int(_positive_number(api_config.get("max_prefetch_tables"), _DEFAULT_MAX_PREFETCH_TABLES))
+        max_tables = coerce_positive_int(api_config.get("max_prefetch_tables"), _DEFAULT_MAX_PREFETCH_TABLES)
         if len(tables) > max_tables:
             return Result(
                 success=False,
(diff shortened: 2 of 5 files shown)


arnozhang:
Both findings confirmed and fixed in 07f8a997. #1 was a real hole in exactly the guarantee the new helper's docstring makes.

**1. `_positive_number` and `.inf` — confirmed, with a second value in the same shape.**

Reproduced each step: `float('.inf')` passes `> 0`, `int(float('inf'))` raises `OverflowError`, and PyYAML parses `.inf` to float infinity. So `max_prefetch_tables: .inf` — the natural way to write "no limit" — 500s every `POST /table/columns`. And `prefetch_budget_seconds: .inf` makes the deadline infinite, silently restoring the unbounded batch this PR exists to remove. `nan` was already safe (`nan > 0` is False).

Your pointer to `chat_service._positive_int` was the right call — it catches `OverflowError` and names the `.inf` case in its docstring. Rather than write a third variant I moved it to `utils/config_utils`, whose module docstring is literally about coercing values that arrive from YAML, and added `coerce_positive_seconds` there for the budget (keeps fractions, rejects infinity via `math.isfinite`). `chat_service`'s private copy is gone, so there's one implementation.

The `0.5` sub-finding is real too and your recommendation handles it: `coerce_positive_int` gets `int(0.5)` = 0, fails `> 0`, falls back — instead of installing a limit of 0 that rejects everything.

Verified the tests actually catch it: with the pre-fix coercion restored, 5 of the new cases fail with `OverflowError: cannot convert float infinity to integer`.

**2. Process-wide `monotonic` patch — confirmed.**

`database_service.time is time` returns True, and inside the old patch `time.monotonic()` read the fake from unrelated call sites; a 4th reading raised `StopIteration`. Went with the injectable-clock direction rather than just `chain`/`repeat`: a `_fake_monotonic` context manager rebinds the module-level `time` name to a stand-in, so the stdlib clock is untouched — verified `time.monotonic is real` inside the fake — and the tail is `repeat()` so an extra reading degrades an assertion instead of raising.

**On your non-defect observation** — you're right that the budget only gates between tables, so the ceiling is 3s *plus one table's round-trip*, not 3s. I've corrected that in the PR description rather than leave the sharper number standing.

Thanks also for pre-verifying the ordering contract, the structlog positional args, and that `DatasourceService` is cached per project so the gate is a real process-wide singleton — that last one is load-bearing and I had not stated it anywhere.

`test_database_service.py` 83 passed; `tests/unit_tests/api` 1364 passed with the same 10 pre-existing failures + 2 errors as clean `main`; `ruff format` + `check` clean.

arnozhang:
4 and 5 fixed in 2348ea3b. Both were in code I added last round, and 5 turned out to have a sibling the review did not reach.

**5 — confirmed, and the fix is half behaviour, half docstring.**

Reproduced: `1.9 → 1`, `2.9 → 2`, `True → 1`, and PyYAML reads `true`/`yes`/`on` all as `True`. The bool half is unambiguous and now rejected in `coerce_positive_int` — a bound of 1 is worse than an absent one because it looks deliberate and, via finding 3, rejects every multi-table batch.

**`coerce_positive_seconds` had the same hole** — `prefetch_budget_seconds: true` bought a 1-second budget. Not in your list; I found it by exercising the pair together after fixing the first. Also rejected now.

The fraction half I fixed in the docstring rather than the behaviour, because rejecting non-integral floats broke `test_fractional_page_size_config_truncates_to_a_bounded_int` in chat_service — and that test is right. `limit` reaches this helper from **API callers**, not only YAML, and there `3.5 → 3` still bounds the request while falling back would hand back a *larger* default than was asked for. Truncation is benign for a page size and harmful only for a rejection threshold, so the policy belongs at the call site, not in the shared helper. The docstring now says so, including who must check before calling. Which means the `max_prefetch_tables: 1.9 → 1` case survives — and its real fix is finding 3, see below.

**4 — confirmed as latent.** Mechanism holds; this platform's monotonic resolution is 41.7ns so it passes, and there is no Windows runner in CI, so it was never breaking. Swapped both tests onto the deterministic `_fake_monotonic` anyway, with a comment naming the coarse-clock reason.

**Something worth flagging about my own tests from last round.** The `yaml-true` and `fraction-above-one` cases I first added at service level did not catch anything: with a single-table request, a bound of 1 and a bound of 500 are indistinguishable, so they passed with the guards removed. Mutation check caught it. The coercion semantics now live in `tests/unit_tests/utils/test_config_utils.py` next to the existing `coerce_bool` tests, and the service-level parametrize sends **two** tables and keeps only values whose effect is observable through the endpoint. Guards removed → 3 targeted failures; restored → green.

**1, 2, 3 — agreed, follow-ups, and now written into the PR description** under "Known limits — this is not blanket rate limiting", so nobody reads the gate as global. Your `db_manager.py:193` confirmation went in verbatim; I also noted that a module-level gate would be the wrong fix since it would serialize projects on unrelated datasources — a registry keyed by datasource URI is the shape.

On 3 specifically: I think it is the root cause of the whole "silently tiny bound" family — with truncation, a bogus bound of 1 degrades to warming one table per batch instead of rejecting everything, which would have made the fraction half of 5 harmless too. I left it out because it changes a documented response and the test pinning it, which deserves its own review rather than riding along here. Happy to do it next if you agree it is in scope.

`tests/unit_tests/api` + `tests/unit_tests/utils` — 2919 passed, against 2877 on the same tree with my changes stashed; the 11 failures + 5 errors are identical in both runs. `ruff format` + `check` clean.

PR #1371, opened 2026-08-26
[BugFix] Reset the occupancy meter after a major compact

Engmelli:
## Why

`_decide_compact_mode` reads context usage from `_history_token_ratio_sync`, which
prefers the `running_turn_usage` snapshot. `_major_compact` rewrites the history into
a single recap but never touches that snapshot.

So the moment a compact finishes, the meter still reports the pre-compact figure. The
next pre-turn check reads the same near-limit ratio, decides "major" again, and
compacts again — one LLM summarization per turn for the rest of the session, each
discarding more history than the last. We watched a session compact six times in about
two minutes before it was gutted.

## Solution

Reset the meter to the size of the recap after a successful compact, since the recap is
now the whole of the history.

Key decisions:

- **In-memory mirrors are set before `persist_context_state`, not through it.** That
  method returns early when there is no resolvable state path and syncs its mirrors only
  on a successful save, so routing through it would leave `_restored_context_used` stale.
  That matters because at `pre_user_turn` the snapshot has already been dropped at turn
  end, making the restored value the one the ratio actually reads.
- **Only `session_total_tokens` is written.** The ratio reads
  `session_total_tokens or input_tokens`, so the former is the field that matters.
  `input_tokens` is the turn-cumulative counter reported verbatim in the API's end
  event, where a post-compact value would contradict the turn's own output totals.
- **The provider's token count is normalized where it enters.** It can be absent,
  non-numeric or negative, and everything downstream assumes an int — including the `%d`
  in the completion log, which raised `TypeError` on a string. All of it runs after the
  compact has persisted, where raising turns a good compact into a failure.
- **A zero count falls back to measuring the recap text**, reusing the
  4-chars-per-token estimate `compress_utils` already falls back to. Zero is what any
  provider that does not report usage returns, and it would fall through the `tok > 0`
  guard straight back to the stale value.

Tradeoff, deliberate: the recap's token count under-reports true occupancy, which also
includes the system prompt and tool schemas. The bar dips low and jumps on the next
call. Safe as designed — the next call refreshes the snapshot, and a genuine crossing
still fires.

## Test Cases

Nine cases in `tests/unit_tests/agent/node/test_agentic_node.py`. All fail on the
pre-fix code, verified by reverting the source with `git stash push -- <paths>`:

Driving `_major_compact`, so the wiring is what is pinned:
- `test_major_compact_drops_the_ratio_below_the_threshold` — deleting the
  `_reset_context_occupancy` call fails this with `assert 0.95 < 0.8`
- `test_major_compact_leaves_the_turn_cumulative_counter_alone`
- `test_major_compact_measures_the_continuation_when_usage_is_zero` — asserts the
  occupancy matches the continuation actually handed to `session.add_items`
- `test_major_compact_survives_a_provider_reporting_junk_usage`

On the helper:
- `test_meter_is_reset_to_the_recap_size`
- `test_the_ratio_drops_below_the_threshold_after_a_compact`
- `test_a_provider_that_reports_no_usage_still_clears_the_meter`
- `test_the_restored_mirror_is_set_when_there_is_no_snapshot`
- `test_the_ratio_is_safe_even_when_persistence_fails`

Tests use the real `TokenUsage`, so a future `frozen=True` or `validate_assignment=True`
surfaces here rather than degrading into a logged warning.

### DIFF
--- a/datus/agent/node/agentic_node.py
+++ b/datus/agent/node/agentic_node.py
@@ -1907,6 +1907,49 @@ class AgenticNode(Node):
             return None
         return path
 
+    def _reset_context_occupancy(self, summary_token: int) -> None:
+        """Point the occupancy meter at the post-compact history size.
+
+        ``_decide_compact_mode`` reads the ratio from
+        ``_history_token_ratio_sync``, which prefers the ``running_turn_usage``
+        snapshot. Compaction rewrites the history but leaves that snapshot at
+        its pre-compact value, so the next pre-turn check sees the same
+        near-limit figure and compacts again — every turn, until the session is
+        gutted.
+
+        The recap is the whole of the new history, so its token count is the new
+        occupancy. Underestimating is safe: the next LLM call refreshes the
+        snapshot with the true figure, and a genuine crossing still triggers.
+
+        Best-effort throughout — the compact has already succeeded by the time
+        this runs, so nothing here may turn it into a reported failure.
+        """
+        post_compact_tokens = int(summary_token or 0) or 1
+        try:
+            context_length = (
+                int(getattr(self, "context_length", 0) or 0)
+                or int(getattr(self, "_restored_context_length", 0) or 0)
+                or 0
+            )
+        except Exception:  # noqa: BLE001 - a model without a known window
+            context_length = 0
+
+        usage = getattr(self, "running_turn_usage", None)
+        if usage is not None:
+            try:
+                usage.session_total_tokens = post_compact_tokens
+                usage.input_tokens = post_compact_tokens
+            except Exception as exc:  # noqa: BLE001
+                logger.warning("Could not reset running_turn_usage after compact: %s", exc)
+
+        try:
+            if context_length:
+                self.persist_context_state(post_compact_tokens, context_length)
+            else:
+                self._restored_context_used = post_compact_tokens
+        except Exception as exc:  # noqa: BLE001
+            logger.warning("Could not persist post-compact context state: %s", exc)
+
     async def _major_compact(self, *, reason: str) -> Dict[str, Any]:
         """LLM-driven full-history compact pass.
 
@@ -1986,6 +2029,7 @@ class AgenticNode(Node):
         # a single assistant continuation message, so the next minor pass
         # starts scanning from the top of the rewritten session.
         self._compacted_until = 0
+        self._reset_context_occupancy(summary_token)
 
         logger.info(
             "Major compact complete: %d chars summary, %d output tokens, history=%s",
--- a/datus/agent/node/agentic_node.py
+++ b/datus/agent/node/agentic_node.py
@@ -53,6 +53,17 @@ if TYPE_CHECKING:
 logger = get_logger(__name__)
 
 
+def _coerce_token_count(value: Any) -> int:
+    """A provider-reported token count as a non-negative int, or 0.
+
+    Usage dicts come from the provider, so a count may be missing, a string, or
+    negative. Callers run on paths where raising is not acceptable."""
+    try:
+        return max(0, int(value or 0))
+    except (TypeError, ValueError):
+        return 0
+
+
 def _estimate_tokens(text: str) -> int:
     """Rough token count for text, matching the fallback used elsewhere in the
     codebase (see ``compress_utils.count_tokens``): roughly 4 characters per
@@ -1932,13 +1943,12 @@ class AgenticNode(Node):
         Best-effort throughout — the compact has already succeeded by the time
         this runs, so nothing here may turn it into a reported failure.
         """
-        # ``summary_token`` is the recap's reported output tokens, which is 0 on
-        # any provider that does not report usage. A zero would fall through the
-        # ``tok > 0`` guard in ``_history_token_ratio_sync``, sending it back to
-        # the stale pre-compact fallback — the loop this exists to close. Fall
-        # back to a size estimate of the text actually persisted, so those
-        # providers get a real floor rather than a sentinel.
-        post_compact_tokens = int(summary_token or 0) or _estimate_tokens(continuation) or 1
+        # A zero — any provider that does not report usage — would fall through
+        # the ``tok > 0`` guard in ``_history_token_ratio_sync`` and send it back
+        # to the stale pre-compact fallback, which is the loop this exists to
+        # close. Measure the text actually persisted instead, so those providers
+        # get a real floor rather than a sentinel.
+        post_compact_tokens = _coerce_token_count(summary_token) or _estimate_tokens(continuation) or 1
         try:
             context_length = (
                 int(getattr(self, "context_length", 0) or 0)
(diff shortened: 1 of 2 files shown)


liuyufei9527 (commented):
## Review

The fix is correct and lands in the right place. I verified the premise holds: `compact()` has no cooldown and `_decide_compact_mode` gates major solely on the ratio, so nothing else would have broken the loop. One real gap though — **none of the five new tests pins the call site**. I replaced `self._reset_context_occupancy(summary_token)` with `pass` on your branch and all 132 tests in `test_agentic_node.py` still passed.

### What I verified as correct

- **Call placement.** After the summarization LLM call (whose own `TokenUsageHook` writes the pre-compact figure), and only on the success path — a failed compact can't masquerade as a shrunk history.
- **Both trigger paths covered.** `hook_major` goes through the in-memory snapshot; at `pre_user_turn` the snapshot has already been dropped at turn end (`agentic_node.py:3284`), so the `_restored_context_used` fallback is what matters. Both are written.
- **Ordering.** Setting the in-memory mirrors before `persist_context_state` is right, and the comment explains why — `persist_context_state` returns early with no resolvable state path and only syncs its mirrors on success.
- **Leaving the SQLite `running_turn_usage` row stale is correct.** It only feeds `get_detailed_usage`'s cumulative totals, where those tokens really were spent, and no production code reads it back into the ratio path.
- Defensive `getattr` on `_restored_context_length` is warranted — `__init__`-bypassed nodes don't set it.
- `ruff format --check` and `ruff check` clean; `ci/audit_tests.py --diff-only` reports P0=0 / P1=0; the 17 compact-related tests pass.

### 1. No test pins the call site (main ask)

All five tests call `_reset_context_occupancy` directly. They fail on pre-fix code only because the method doesn't exist (`AttributeError`) — the weakest form of red. The regression this PR fixes is not actually held down: delete the one line in `_major_compact` and the suite stays green.

There's a ready-made slot next to `test_major_compact_success_persists_continuation_message` (`tests/unit_tests/agent/node/test_agentic_node.py:820`), which already drives the full `_major_compact`. I ran this and confirmed it passes on your HEAD and goes red when the call site is removed:

```python
@pytest.mark.asyncio
async def test_major_compact_drops_the_ratio_below_the_threshold(self):
    node = _make_node(context_length=100_000)
    node.session_id = "compact_test"
    node._session = _make_async_session_mock()
    node._session_manager = MagicMock()
    mock_model = MagicMock()
    mock_model.context_length.return_value = 100_000
    mock_model.generate_with_tools = AsyncMock(
        return_value={"content": "recap", "usage": {"output_tokens": 100}}
    )
    node.model = mock_model
    node.running_turn_usage = TokenUsage(
        session_total_tokens=95_000, input_tokens=95_000, context_length=100_000
    )
    assert node._history_token_ratio_sync() >= 0.8

    with patch.object(_ConcreteAgenticNode, "_dump_session_history_jsonl", new=AsyncMock(return_value=None)):
        with patch.object(_ConcreteAgenticNode, "_get_archive", return_value=None):
            with patch.object(_ConcreteAgenticNode, "_get_system_prompt", return_value="sys"):
                result = await node._major_compact(reason="hook_major")

    assert result["success"] is True
    assert node._history_token_ratio_sync() < 0.8
```

### 2. `SimpleNamespace` stands in for `TokenUsage`

`TokenUsage` is already imported at line 25 of that test file. The production write is wrapped in `try/except` + `logger.warning`, so if `TokenUsage` ever gains `frozen=True` or `validate_assignment=True`, the fix degrades into a silent warning and all five tests still pass. CLAUDE.md asks for cross-component contracts to consume the producer's real output — at least one case should use the real model (the snippet above does).

### 3. `usage.input_tokens = post_compact_tokens` is unnecessary and lossy

`_history_token_ratio_sync` reads `session_total_tokens or input_tokens`, and `session_total_tokens` is guaranteed truthy after the `or 1`, so `input_tokens` is never consulted on this path. Meanwhile `input_tokens` is the turn-cumulative counter (`TokenUsageHook._publish` feeds it from the SDK accumulator) and is forwarded verbatim to the API's SSE `end` event (`datus/api/services/chat_task_manager.py:882`). If no further LLM call happens in the turn after the compact (`reason="overflow"`, or a compact fired on the final tool call), that event reports `input_tokens=1200` alongside an `output_tokens`/`total_tokens` in the tens of thousands — internally inconsistent. Suggest dropping the line, or keeping it with a comment explaining the redundancy.

### 4. The `or 1` sentinel deserves a comment, and the test pins the wrong thing

The reason for `int(summary_token or 0) or 1` — a 0 would fall through the `tok > 0` guard in `_history_token_ratio_sync` and re-read the stale fallback — is in neither the code nor the docstring. And this isn't a hypothetical branch: `summary_token = usage.get("output_tokens", 0)` is genuinely 0 for any model path that doesn't report usage, so on those providers the meter reads ~0% — the largest possible under-report. `len(continuation) // 4` is right there and makes a better floor.

Relatedly, `test_a_zero_token_recap_still_clears_the_meter` asserts `== 1`, pinning the sentinel rather than the property. Per CLAUDE.md ("assert the property, not the instance"), assert the ratio drops below the threshold instead, so a better estimate later doesn't break the test.

### 5. Under-reporting note (acknowledged in the description, not blocking)

`summary_token` is the recap's output tokens, but the next call's real occupancy also includes the system prompt and tool schemas — easily 10k+ in Datus. The context bar will dip low and then jump on the next call. Safe as designed (the next call refreshes, a genuine crossing still fires); the floor in #4 would narrow the gap.

### 6. PR description

The body doesn't use the mandatory template (`## Why` / `## Solution` / `## Test Cases`), which CLAUDE.md marks non-negotiable. The test table is also stale: it lists four cases including `test_a_missing_meter_is_not_an_error`, which doesn't exist on the branch; there are five, two named differently.

### Follow-up (out of scope here)

When a model path leaves `usage.request_usage_entries` empty, `last_call_input_tokens` is 0 (`datus/models/openai_compatible.py:1877`), and `TokenUsageHook._update_node_snapshot` then fills `session_total_tokens` from the monotonically growing turn-cumulative `input_tokens`. On those paths this fix only postpones the loop by one LLM call — `session_total_tokens` doesn't mean "current window occupancy" on that fallback. Worth a separate issue.

Engmelli:
All six addressed, thanks.

liuyufei9527:
Closing this in favour of #1406, which fixes the same re-compact loop by a different and stricter route.

The diagnosis here is right: `_decide_compact_mode` reads the ratio through `_history_token_ratio_sync`, and a major pass used to rewrite the history without touching the measurement behind it, so the next pre-turn check re-triggered on the stale near-limit figure.

Where #1406 differs is what a rewrite leaves behind. Instead of repointing the meter at the recap's token count, `_replace_session_items` calls `_invalidate_context_usage()`, which marks the measurement invalid in memory and in SQLite inside the same transaction. `_history_token_ratio_sync` returns `0.0` whenever `state.valid` is false, so the loop closes just the same, and the next model response supplies a real measurement.

That makes the two approaches mutually exclusive rather than merely overlapping. A core rule of #1406 is that nothing may substitute a text estimate or a cumulative counter for actual input tokens, so `_reset_context_occupancy` writing `len(continuation) // 4` into `session_total_tokens` is exactly the kind of synthetic measurement it removes. `_estimate_tokens` has no role left once occupancy can be explicitly unknown.

Thanks for the write-up and the nine regression tests; the failure analysis in the description is what made the shared root cause easy to confirm.

PR #1372, opened 2026-08-26
[BugFix] Do not run a major compact mid-turn

Engmelli:
## What breaks

A major compact clears the session and writes a recap in its place. Mid-turn,
that achieves nothing.

The in-flight agents-SDK run holds the conversation **in memory** and keeps
sending it. Clearing the session shrinks nothing the model actually receives, so
the occupancy the trigger reads never drops — and the next tool call trips the
same threshold again.

The result is a summarization every few seconds for the remainder of the turn.
We measured roughly **one every 13 seconds** across a heavy turn: each an LLM
call, each changing nothing the model can see.

## The fix

Only the compact at the **start of the next turn** — before the session is
loaded into a run — changes what the model is sent. Mid-turn majors become a
no-op, and the work happens there instead.

Minor compaction is untouched: it was already decided only at turn start.

## The test change

`test_mid_turn_still_allows_major` pinned the old behaviour, so it is updated
rather than added to. Its previous rationale — *"major still fires mid-turn, its
token-ratio gate genuinely changes as the turn progresses"* — is true about the
ratio but not about the outcome: the gate moves, and acting on it still cannot
shrink anything.

The replacement pins both halves of the real contract:

```python
assert await node._decide_compact_mode(mid_turn=True) == "noop"
assert await node._decide_compact_mode() == "major"   # same state, turn start
```

## Verification

- `tests/unit_tests/agent/node`: failure set **identical** to clean `main` — 0
  introduced.
- `ruff check datus/` clean.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
- [ ] release/0.3.8
- [ ] release/0.3.9
- [ ] release/0.4.0

### DIFF
--- a/datus/agent/node/agentic_node.py
+++ b/datus/agent/node/agentic_node.py
@@ -1727,6 +1727,19 @@ class AgenticNode(Node):
         except Exception:
             ratio = 0.0
         if cfg.major.enabled and ratio >= cfg.major.token_threshold:
+            if mid_turn:
+                # A mid-turn major cannot help. The in-flight agents-SDK run
+                # holds the conversation in memory, so clearing the session
+                # shrinks nothing the model is being sent: occupancy does not
+                # drop, and the trigger fires again after the very next tool
+                # call. Only the compact at the START of the next turn, before
+                # the session is loaded into a run, changes what the model sees.
+                logger.info(
+                    "Mid-turn major compact suppressed (ratio=%.2f): the live run's context "
+                    "cannot shrink; compacting at next turn start instead.",
+                    ratio,
+                )
+                return "noop"
             return "major"
         if mid_turn:
             return "noop"
--- a/tests/unit_tests/agent/node/test_compact_helpers.py
+++ b/tests/unit_tests/agent/node/test_compact_helpers.py
@@ -145,12 +145,18 @@ class TestDecideCompactMode:
                 assert await node._decide_compact_mode() == "minor"
 
     @pytest.mark.asyncio
-    async def test_mid_turn_still_allows_major(self, tmp_path):
-        """major still fires mid-turn — its token-ratio gate genuinely changes
-        as the turn progresses."""
+    async def test_mid_turn_major_is_suppressed(self, tmp_path):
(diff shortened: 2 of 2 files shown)


liuyufei9527:
Thanks for the detailed writeup, and for actually measuring the frequency — the diagnosis is correct, and we confirmed it independently.

**Root cause confirmed.** In `openai-agents`, `_prepare_input_with_session()` runs exactly once per run, before the `while True` loop (`run.py:577` for `Runner.run`, `run.py:1145` for `run_streamed`). Inside the loop the SDK works off the in-memory `original_input` / `generated_items` and only ever *writes* to the session via `_save_result_to_session`. So clearing the session mid-run genuinely changes nothing the model receives: occupancy never drops and the trigger re-fires on the next tool call — exactly the loop you measured.

**One more effect worth recording.** The run writes the user message to the session at start (`run.py:1152`), and a mid-turn `clear_session()` deletes it. The loop afterwards only appends `new_step_items`, so that message is never rewritten — the turn's own question is lost from the session, and `_user_turn_count_from_session()` under-counts as a result. That makes the current behaviour worse than just wasted LLM calls.

**Why we're not merging this as-is.** Suppressing the mid-turn major makes `_decide_compact_mode(mid_turn=True)` return `noop` on every path, which leaves `CompactHook.on_tool_end` unreachable in production — the hook, its module docstring, and several of its unit tests would then describe behaviour that can never happen. More importantly, mid-turn compaction is a capability we do want. The bug isn't that it runs mid-turn; it's that it's applied to the session instead of to the outgoing request. The proper fix moves it to the model-request boundary, where trimming actually shrinks what the model is sent, while preserving `tool_call` / `tool_result` pairing.

That's beyond this PR's scope, and it has to land across all three model paths — including the native Anthropic loop, which doesn't go through `Runner` at all.

We're prioritising the proper fix now. Your diagnosis is what got us there, and we'll reference this PR in the fix. I'll close this one once that lands.

PR #1373, opened 2026-08-26
[BugFix] Keep a session's title across a major compact

Engmelli:
## What breaks

The sidebar title comes from scanning a session for its first user message.

A major compact clears the session and re-adds **only an assistant recap**. No
user rows survive, so the scan finds nothing and the chat goes untitled — the
operator can no longer identify it in the session list.

Once the conversation continues, it gets worse rather than better: the earliest
surviving user row is now some mid-conversation message, so the chat silently
**retitles itself** to whatever was said after the compact.

## The fix

A `<session_id>.title` sidecar is written before the history is cleared, and
preferred on read. The title stays the original first message across any number
of compacts.

Two details that matter:

- **It is only ever written when a title can still be read.** A later compact
  sees no user rows, so without that guard it would overwrite a good title with
  nothing.
- **Saving is best-effort.** A lost title must never fail a compact, and the
  call site tolerates a node whose `session_manager` cannot be resolved.

## Tests

Four cases in `TestTitleSidecar`:

| test | asserts |
|---|---|
| `test_title_survives_a_history_that_lost_its_user_rows` | the post-compact shape — assistant recap, no user rows — still reports the original title |
| `test_the_sidecar_wins_over_a_later_user_message` | a mid-conversation user row does not retitle the chat |
| `test_no_sidecar_leaves_the_scan_untouched` | ordinary sessions are unaffected |
| `test_an_empty_title_is_never_written` | a blank save cannot clobber a good title |

Three of the four fail with the change reverted; the fourth is the
untouched-behaviour guard and passes either way by design.

## Verification

- `tests/unit_tests/agent/node` and `tests/unit_tests/models`: failure set
  **identical** to clean `main` — 0 introduced (790 pre-existing in this
  environment).
- `ruff check datus/` clean.
)

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved session titles when conversation history is compacted.
  - Kept sidebar titles stable even when earlier messages are removed or later messages might otherwise change the title.
  - Ignored blank replacement titles to prevent accidental loss of existing session names.

- **Tests**
  - Added coverage for title preservation across compaction and normal session histories.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
- [ ] release/0.3.8
- [ ] release/0.3.9
- [ ] release/0.4.0

### DIFF
--- a/datus/agent/node/agentic_node.py
+++ b/datus/agent/node/agentic_node.py
@@ -1907,6 +1907,24 @@ class AgenticNode(Node):
             return None
         return path
 
+    def _preserve_session_title(self) -> None:
+        """Save the session's current title so a compact cannot erase it.
+
+        Best-effort: a lost title must never fail a compact, and
+        ``session_manager`` is a property that resolves a path manager, so it can
+        raise for a node with no project context.
+        """
+        if not self.session_id:
+            return
+        try:
+            manager = self.session_manager
+            info = manager.get_session_info(self.session_id) or {}
+            title = (info.get("first_user_message") or "").strip()
+            if title:
+                manager.save_title_sidecar(self.session_id, title)
+        except Exception as exc:  # noqa: BLE001
+            logger.warning("Could not preserve the session title before compacting: %s", exc)
+
     async def _major_compact(self, *, reason: str) -> Dict[str, Any]:
         """LLM-driven full-history compact pass.
 
@@ -1935,6 +1953,10 @@ class AgenticNode(Node):
             return {"mode": "major", "reason": reason, "success": False, "summary": "", "summary_token": 0}
 
         logger.info("Starting major compact for session %s (reason=%s)", self.session_id, reason)
+        # Capture the chat's title BEFORE the history is cleared: the compact
+        # re-adds only an assistant recap, leaving no user rows for the sidebar's
+        # first-user-message scan to find.
+        self._preserve_session_title()
         history_jsonl_path = await self._dump_session_history_jsonl()
 
         sys_prompt = self._get_system_prompt()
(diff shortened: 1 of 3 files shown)


liuyufei9527:
Closing in favour of #1407, which fixes this bug by a different route. The
problem analysis here was right and carried over directly, so thank you for it
— both failure modes you identified are the ones #1407 pins down: a compacted
session reporting no title at all, and the worse case where the conversation
continues and the chat silently retitles itself to a mid-conversation message.

Two things changed in the implementation.

**Where it is stored.** #1407 puts the title in a `session_meta` table inside
the session database rather than a `<session_id>.title` sidecar. That makes
`delete_session` collect it with the database file (no separate cleanup, and no
user text left on disk after a delete), gives it the same path validation as
every other session file, and makes the write atomic through the SQLite
transaction. It also lets `copy_session` and `rewind_session` carry the row
over — without that, switching node type or rewinding a compacted session
produces an untitled session again one step later, which the sidecar version
would still have hit.

**When it is written.** Instead of salvaging the title before a compact clears
the history, it is recorded as the first user message arrives, with
`INSERT OR IGNORE` so only the first one ever wins. That removes the timing
problem rather than working around it: nothing races the clear, and the guard
you needed against a later compact overwriting a good title with nothing is no
longer necessary because a later write simply cannot take effect.

Your test cases carried over too, in the form of the `session_meta`-backed
equivalents — the post-compact shape with no user rows, and a mid-conversation
user message not retitling the chat. Both were verified red against the
pre-fix code.

PR #1434, opened 2026-09-17
[BugFix] interaction: stop an unanswered prompt from holding the turn open

arnozhang:
## Why

`InteractionBroker.request()` awaited its future with no bound:

```python
result = await future   # no timeout
```

That await is **inline** in the tool that raised the prompt, the permission-prompt lock is held across it, and the parent turn only finishes once every tool call in it returns. So one card nobody answers freezes the whole turn, `ActionBus.merge` never drains, and the SSE stream never reaches its `end` event.

A customer session hit exactly that (`chat_session_1da22411`, trace `14490984b564443f9fff3fea6e9ecbe6`):

| time | event |
|---|---|
| 07:25:11 | main agent dispatches two `explore` sub-agents in parallel |
| 07:25:34 | sub-agent #2 calls `bash`; `bash_tools.bash -> ask` → `InteractionBroker: request queued with action_id=970602a8-…` |
| 07:25:34 → 07:31:53 | sub-agent #1 streams **2649 more frames / 95 bubbles**, carrying the card off screen |
| 07:31:53 → 21:00:33 | **13.5 hours of zero application logs** — only IDE WebSocket ping/pong |
| 21:00:33 | reader closes the tab → `request_completed duration_ms=49042966` (13.6h) |

`POST /api/v1/chat/user_interaction` was never called. The SSE heartbeat kept the connection healthy the whole time, so the UI never errored or timed out — it just spun.

## Solution

**Bound the wait.** Default 30 minutes — well past any realistic "went to get coffee" gap, still finite. `DATUS_INTERACTION_TIMEOUT_SECONDS` overrides it; `0` restores the old unbounded behaviour for anyone who wants it.

**Raise `InteractionTimeout(InteractionCancelled)`.** All four call sites (`_request_user_confirmation`, `_request_external_confirmation`, `_handle_sql_permission`, `confirm_plan`, `ask_user`) already have `except InteractionCancelled` taking the safe branch — deny, report unconfirmed, give up. Inheriting reuses that instead of adding a fourth code path each caller has to remember.

**A timeout never resolves to the event's `default_choice`.** Tempting, and correct for a permission prompt (`default_choice="n"` = deny), but `confirm_plan` sets `default_choice="confirm"` — so the "obvious" version would silently auto-confirm a plan nobody read. Hanging is bad; confirming unread work is worse. There is a test pinning this.

**Queue a notice.** A `role=INTERACTION, status=SUCCESS` action with `output["content"]` set, which `_build_interaction_result_content` already renders as markdown. Without it the card just stops mattering with nothing said and the transcript reads as if the agent ignored its own question.

**`ask_user` distinguishes the two cases.** Told the user "cancelled", the model reads a decision and asks again; told nobody answered, it proceeds on a stated assumption.

## Test Cases

`tests/unit_tests/cli/test_execution_state.py::TestInteractionRequestTimeout` — 7 new cases:

- an unanswered request raises `InteractionTimeout` and clears `_pending`
- the timeout is catchable as `InteractionCancelled` (the inheritance the four call sites rely on)
- **a timeout never resolves to `default_choice`**, pinned with the `confirm_plan`-shaped prompt
- a client-visible notice is queued, with non-empty `output["content"]` (an empty one is dropped by the converter and the transcript would say nothing)
- an answer inside the window still wins
- `timeout <= 0` waits forever
- `DATUS_INTERACTION_TIMEOUT_SECONDS` parsing: override, `0` → disabled, garbage → default, unset → default

`pytest tests/unit_tests/cli/test_execution_state.py tests/unit_tests/tools/func_tool/test_ask_user_tools.py tests/unit_tests/tools/permission/test_permission_hooks.py` → 308 passed.

## Notes for the reviewer

This is the backstop, not the fix for the root UX problem — the reader could not *see* the card. The companion Datus-saas PR (#1020) adds a bar above the composer that counts pending cards which are off screen and scrolls to them, and makes the streaming indicator say "waiting for your confirmation" instead of rotating "Thinking…". This PR makes sure that even when nobody answers, the run ends.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
- [ ] release/0.3.8
- [ ] release/0.3.9
- [ ] release/0.4.0
- [ ] release/0.4.1
## Summary by CodeRabbit

* **New Features**
  * Added configurable time limits for unanswered interactions, with a 30-minute default.
  * Added support for overriding the timeout per request or through an environment setting.
  * Added timeout notifications when an interaction expires.
  * Timed-out questions now provide guidance to continue with a reasonable default.

* **Bug Fixes**
  * Pending interaction requests are cleaned up after timing out.
  * Explicitly disabled timeouts continue to allow unlimited waiting.

## Summary by CodeRabbit

* **New Features**
  * Added configurable time limits for unanswered interactions, with a 30-minute default.
  * Added support for overriding timeouts per request or through an environment setting.
  * Added timeout notifications when an interaction expires.
  * Timed-out questions now provide guidance to continue with a reasonable default.

* **Bug Fixes**
  * Pending interaction requests are cleaned up after timing out.
  * Answers submitted as a timeout occurs are handled correctly.
  * Explicitly disabled timeouts continue to allow unlimited waiting.
  * Invalid timeout settings safely fall back to the default.
  * Timed-out interaction choices are no longer offered to clients.


arnozhang:
## Verified end-to-end in a browser, not just in unit tests

Ran the companion `/saas-verify` suite (Datus-ai/Datus-saas#1020) against a local `datus_backend` built from this branch, started with `DATUS_INTERACTION_TIMEOUT_SECONDS=75` so the wait could be observed inside one run.

Real chat turn, real `bash_tools.bash -> ask`, card left deliberately unanswered:

```
04:29:42  Permission check: bash_tools.bash @ chat = ask
04:29:42  InteractionBroker: request queued with action_id=a225e0c8-3c7b-4528-9bc2-ada178c2d2ff
04:30:57  InteractionBroker: action_id=a225e0c8-... went unanswered for 75s;
          giving up so the run can finish instead of holding the turn open
04:30:57  chat interaction failed: Error running tool bash:
          User rejected execution of bash command
```

The browser side confirmed what the log implies: `the turn ends on its own instead of hanging — streaming stopped after ~75s`, the SSE stream reached its `end` event, the card rendered as expired with no buttons, and the transcript carried the "No answer within N min" notice. Before this change that same shape ran for 13.6 hours.

The safe-branch inheritance was exercised too: `InteractionTimeout` was caught by the existing `except InteractionCancelled` in the permission hook and reported to the model as a rejection — the tool result above is that path, unmodified.

The answered path is unaffected — the same run answered an earlier card normally (`InteractionBroker: received response for action_id=...: [['y']]`) and the turn continued.

arnozhang:
Review follow-ups pushed in `adcc0ac`. Two of them are worth spelling out because the obvious fix was wrong.

**The timeout notice could not retire the card it killed.** `timed_out_action_id` was in the action's `input`, but `_build_interaction_result_content` forwards only `output`, and the notice carries a fresh uuid rather than the dead card's id. Crucially the turn does *not* have to end with the timeout — a sibling sub-agent streams on — so the client had no way to distinguish that card from a live one: it kept its buttons, and pressing one earned "unknown action_id". That is precisely the shape the companion frontend PR exists to remove, reopened here at the moment a reader is most likely to hit it. The id now travels in `output` and the converter emits it as `timedOutInteractionKey`; Datus-ai/Datus-saas#1020 consumes it and retires the card mid-turn.

**The submit/timeout race needed a handshake, not a guard.** My first attempt checked `future.done() && !future.cancelled()` in the timeout handler. That cannot work: `wait_for` cancels the future *before* raising `TimeoutError`, so by the time the handler runs the future is cancelled and the answer is unreachable through it. `submit` now parks the answers in a `_settled` map inside the same critical section as the `_pending` pop, and the timeout path honours anything it finds there. The map is consumed on both the answered and timed-out paths and cleared by `close()`.

The test for it is deterministic rather than wall-clock: it hands the answer over exactly as `submit` does and never resolves the future, so `wait_for` is guaranteed to time out and the assertion is on which side wins, not on scheduling luck. A first version raced two 0.05s deadlines and failed about as often as it passed — not a test worth having.

Also in this commit: `nan` / `inf` rejected in the env parse (both silently removed the bound), and the constant now states that 30 minutes is the budget for **one card**, not for the turn — `PermissionHooks` serialises prompts on a per-broker lock held across `request()`, so N sub-agents each needing a permission wait in turn and the worst case is N × the value. The incident that motivated this had two, so an hour. Raising it to a turn-level deadline is a reasonable follow-up; I did not want to change the shape of the lock in a fix whose job is to stop an unbounded wait.

Corrected the PR body's "four call sites" → five.

PR #1438, opened 2026-09-18
[BugFix] Stop an unbound per-request datasource from killing the chat turn

arnozhang:
## Why

A customer hit this on every chat: the first message of a conversation answered
normally, every message after it failed with

```
error_code=100004, error_message=Unsupported value `__SR` for field `datasource`
```

`StreamChatInput.datasource` is a UI hint — the web composer keeps its own
datasource picker selection and sends it with each turn — but `start_chat`
assigned it straight to `AgentConfig.current_datasource`, whose setter raises
`COMMON_UNSUPPORTED` for any name the project has not bound.
`ChatService.stream_chat` turns that into an `error` SSE event, so the whole
turn dies before the agent runs.

The "first one works" part is the picker: `initSessionDatabase` resolves
asynchronously, so the first turn is sent with no datasource at all (the project
default is used) and from the second on it carries a selection that — in the
reported case — belonged to another project. `__SR` is the namespace form of a
Chinese datasource name (`_sanitize_namespace` in Datus-backend maps every
non-`[a-zA-Z0-9_-]` character to `_`), i.e. a datasource that exists in the
account but is not bound to this project.

saas stops sending the stale name (Datus-ai/Datus-saas#1031), but a stale hint
must not be fatal whatever sends it — an IM channel pins a datasource too, and
any of them can be unbound while a session is open.

## Solution

In `ChatTaskManager.start_chat`, apply the per-request `datasource` only when the
project actually binds it; otherwise log a warning naming the value and the
available datasources, and leave the clone on the project's default (the request
field is rewritten to match so downstream context agrees).

`AgentConfig.current_datasource` keeps its strict setter — config assembly
should still fail loudly on a name that does not exist; this is only about the
per-request override.

## Test Cases

`tests/unit_tests/api/services/test_chat_task_manager.py`:

- `TestStartChatDatasourceOverride::test_unbound_request_datasource_falls_back` —
  replaces `test_invalid_request_datasource_raises`, which asserted exactly the
  behaviour this PR removes. Asserts the clone stays on the project's default and
  the turn starts.
- `test_request_datasource_switches_current_datasource` (existing) still covers
  the bound case.

Run: `pytest tests/unit_tests/api/services/test_chat_task_manager.py` → 147 passed.

## Backport

- [ ] release/0.3.1
- [ ] release/0.3.2
- [ ] release/0.3.3
- [ ] release/0.3.4
- [ ] release/0.3.5
- [ ] release/0.3.6
- [ ] release/0.3.7
- [ ] release/0.3.8
- [ ] release/0.3.9
- [ ] release/0.4.0
- [ ] release/0.4.1
## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid per-request datasource names now safely fall back to the project’s default datasource instead of interrupting the chat turn with an error.
  * Valid datasource overrides continue to work as expected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid per-request datasource names now safely fall back to the project’s default datasource instead of interrupting the chat turn.
  * Catalog, database, and schema details are refreshed to match the datasource actually in use.
  * Valid datasource overrides continue to preserve their requested database and schema settings.
  * Repeated requests using an unbound datasource now produce a single warning per chat session, avoiding duplicate notifications.


arnozhang:
Review follow-up pushed (3d1cecd6).

**The stale selection is dropped whole, not datasource-only.** The first commit
rewrote `request.datasource`, which — as pointed out — nothing reads after that
line, while `request.catalog` / `request.database` / `request.db_schema` *do*
reach the agent via `_create_node_input` (chat_task_manager.py:726-728, 909-911).
`_fill_database_context` prefers the request's value over the config's, so the
fallback switched the connection to the project's default datasource and then
pointed it at the dropped datasource's database — another warehouse's same-named
table, silently. All three are now cleared in the fallback branch, so the context
is refilled from the datasource actually in use.

**The warning is now once per (session, datasource)**, not once per turn: an IM
channel pinned to a datasource that was later unbound repeats the same name
forever, and only the first one is news. Bounded at 512 entries (cleared
wholesale when full — forgetting costs one extra log line).

New tests in `TestStartChatDatasourceOverride`:
- `test_unbound_request_datasource_drops_its_database_context` — catalog/schema
  come back `None` and `database` is refilled from the datasource in use, not
  from the request.
- `test_bound_request_datasource_keeps_its_database_context` — a selection that
  still resolves is passed through untouched.
- `test_unbound_datasource_warns_once_per_session` — three turns, one warning.

`pytest tests/unit_tests/api/services/` → 1047 passed, 10 failed; all 10 are
`test_explorer_service.py::TestExplorerServiceOSIAuthoring` and fail identically
on a clean `upstream/main` worktree (verified), so they predate this branch.

Still open, deliberately not in this PR: the fallback is invisible to the user —
they picked `__SR` in the composer and the answer now comes from another
warehouse with nothing on screen saying so. An SSE `notice` (or something the
client can render) is the real fix; filed here rather than bolted on, since it
needs a client-side counterpart.

